PR #1, opened 2023-05-08
Ensure "Create Mission" and "Create Event" buttons appear immediately after login

colincwilliams-kcsara:
The "Create Mission" and "Create Event" buttons were not showing up immediately after logging in. Refreshing the page via the browser resulted in them showing up.

Root caused this to the fact the organization data, which dictates whether those buttons are allowed to show, is exclusively calculated server-side (never recalculated client-side) _and_ is only sent to the client if there's already a valid user.

Looking through the code and database structure I saw no dependency between org and user data. Thus, I opted to fix this bug by always sending the org data to client rather than just when there's a user.

If that gating needs to remain for reasons I didn't see, I'm happy to explore other options. Most likely will mean making an API call to the server to get the org data after the user is logged in, and clearing that data on logout.

### DIFF
--- a/src/app/LoginPanel.tsx
+++ b/src/app/LoginPanel.tsx
@@ -5,6 +5,8 @@ import { CredentialResponse, GoogleLogin, GoogleOAuthProvider } from '@react-oau
 import Api from '@respond/lib/api';
 import { useAppDispatch, useAppSelector } from '@respond/lib/client/store';
 import { AuthActions } from '@respond/lib/client/store/auth';
+import { OrgActions } from '@respond/lib/client/store/organization'
+import { AuthResponse } from '@respond/types/authResponse'
 
 export default function LoginPanel() {
   const { noExternalNetwork } = useAppSelector(state => state.config.dev );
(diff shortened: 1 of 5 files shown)


colincwilliams-kcsara:
I should also note that I smoke tested the app after making this change - logged in, created a mission, updated my status on the mission. Saw no errors or anomalies.

mcosand:
There was a goal here, to not expose all of the org details until the user had authenticated. Some of this was cobbled together as I was trying to figure out Next.js, so something didn't get finished, or broke. I don't know if partnerships are secrets we need to keep from everyone - punt it for now, or try to figure out how to give auth'd users more info about their organization?

colincwilliams-kcsara:
Gotcha, that makes sense. I doubt it would be a problem to expose org details, but agreed with erring on the side of caution.

I'll explore getting org data to authenticated (and only authenticated) users, then. Shouldn't be overly difficult, and I think this is a pretty critical bug for MVP. If we ship with this we'll hit a fair amount of confusion potentially right as a mission is spinning up.

colincwilliams-kcsara:
Updated to send the organization info back with the auth response and save it there. Also clear the organization info on logout.

PR #2, opened 2023-05-09
Complete/Reactivate Activities

colincwilliams-kcsara:
Adds support for marking an activity complete and reactivating it. An activity is complete when it has an end time.

Notes and potential discussion points:
- I used the terminology "Active/Complete" (and variants) as I felt it fits better with SAR's vernacular. Not married to it and can switch to "Open/Closed" in code, the UI, or both if preferred.
- I am _always_ marking all participants as clear when completing an activity. It feels like an invalid state to have a completed activity with non-clear participants, so this prevents getting into that state.
- In the activities list, there will always be completed activities shown, regardless of how many active there are. This was deliberate as, if all complete activities were pushed off the end due to the number of active, it would be impossible to reactivate one that was accidentally completed until one or more other activities completed. Once the overflow page exists this becomes a non-issue.
- To update all of the participants on completion, I'm calling the 'participant/update' reducer from the 'activity/complete' reducer. This feels a bit weird, and I'm open to other approaches. It does seem to follow the [Collection / Item Reducer Pattern](https://redux.js.org/usage/structuring-reducers/reusing-reducer-logic#collection--item-reducer-pattern), though.

Adds the following:
- Button on the activity page to complete/reopoen
- Adds end time to activity page for completed activities
- Confirmation dialog before completing/reopening
- When completing clears all participants
- Hides participant status update button on main page and activity page for completed activities
- Updates the activity lists on the main page:
  - Sorts activities by active then start time (most recent first)
  - Limits the number of completed activities visible to 3 (happy to change this if that's too low)

### DIFF
--- a/src/lib/state/activityReducers.ts
+++ b/src/lib/state/activityReducers.ts
@@ -35,6 +35,7 @@ export const BasicReducers: ActivityReducers = {
   [ActivityActions.complete.type]: (state, { payload }) => {
     const activity = state.list.find(f => f.id === payload.id);
     if (activity) {
+      // First set the end time
       activity.endTime = payload.endTime;
 
       // Then clear every participant
--- a/src/lib/state/activityReducers.ts
+++ b/src/lib/state/activityReducers.ts
@@ -25,6 +25,34 @@ export const BasicReducers: ActivityReducers = {
     state.list = state.list.filter(f => f.id !== payload.id)
   },
 
+  [ActivityActions.reactivate.type]: (state, { payload }) => {
+    const activity = state.list.find(f => f.id === payload.id);
+    if (activity) {
+      activity.endTime = null;
+    }
+  },
+
+  [ActivityActions.complete.type]: (state, { payload }) => {
+    const activity = state.list.find(f => f.id === payload.id);
+    if (activity) {
+      activity.endTime = payload.endTime;
+
+      // Then clear every participant
+      for (const pId in activity.participants) {
+        const participant = activity.participants[pId]
+        BasicReducers['participant/update'](state, {
+          payload: {
+            activityId: activity.id,
+            participant: { ...participant },
+            update: {
+              time: payload.endTime,
+              status: ResponderStatus.Cleared,
+            }
+          }});
+      }
+    }
+  },
+
   [ActivityActions.appendOrganizationTimeline.type]: (state, { payload }) => {
     const activity = state.list.find(f => f.id === payload.activityId);
     if (activity) {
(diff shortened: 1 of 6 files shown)


src/lib/state/activityReducers.ts
```diff
@@ -25,6 +25,35 @@ export const BasicReducers: ActivityReducers = {
...
+        const participant = activity.participants[pId]
+        BasicReducers['participant/update'](state, {
+          payload: {
+            activityId: activity.id,
+            participant: { ...participant },
+            update: {
+              time: payload.endTime,
+              status: ResponderStatus.Cleared,
+            }
+          }});
```
mcosand, line 52: I was going to say this might be cleaner as something like
```
const action = ActivityActions.participantUpdate(...);
BasicReducers[action.type](state, action);
```
but the activity creator doesn't take a Participant, so you have to blow it all apart. Maybe we want to change `participantUpdate` to take a Participant, and current callers pass firstname/etc as one object?

src/types/activity.ts
```diff
@@ -58,12 +58,13 @@ export interface Activity {
   isMission: boolean;
   asMission: boolean;
   startTime: number;
+  endTime: number | null;
```
mcosand, line 61: Javascript `null` vs `undefined` is going to result in a land war somewhere some day.

If the object may be instantiated without the property, I'd prefer to use the syntax:
```suggestion
  endTime?: number;

```

mcosand (commented):
A couple of comments, probably not ship stoppers.

src/types/activity.ts
colincwilliams-kcsara: Huh. I swear I initially did that, using undefined rather than null, but it caused issues. Of course I try it now and it works a treat. Agreed that's preferred, so changing to that.

src/lib/state/activityReducers.ts
colincwilliams-kcsara: That still feels wrong to me because callers would be forced to update the timeline list themselves before they call the action, so they could have the payload prepped. But that timeline update strongly feels like it should be done in a reducer.

I can't help but wonder if this is a code smell that something is wrong with the data and/or reducer structure. Do we need to flatten the structure a bit?

Proposal: check this in as-is since it does work, but I'll play around with some refactoring options. I can also play around with it more before checking it in.

A couple options I'm thinking of:
- Pull `participants` up to their own top-level slice, alongside `activities`. This has already started to happen with the action names, e.g. `participant/update` rather than `activity/updateParticipant`.
- Keep participants where they are, but pull the timeline up to be a sibling of participants rather than a child.

mcosand:
[Complete/Reactivate mission](https://trello.com/c/S6wWZUGC/9-complete-reactivate-mission)

PR #4, opened 2023-06-02
Move activity-level buttons away from responders status

colincwilliams-kcsara:
To reduce confusion, move the buttons that control an activity (edit, complete/reactivate, delete) away from the responder status button. Specifically to the right of the title.

Validated visual behavior with long titles.

Note that the height needed to be set for the button stack, otherwise the buttons would expand in height as the title wrapped onto multiple lines. The fixed height matches the font size of the title so that the buttons are as tall as one line of title text.

### DIFF
--- a/src/app/(main)/EventPage.tsx
+++ b/src/app/(main)/EventPage.tsx
@@ -101,19 +101,25 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
     const isActivityActive = isActive(activity)
     body = (
       <Box>
-        <Typography variant="h4">{activity.title}</Typography>
+        <Stack direction="row" spacing={1} sx={{mt:2, mb:2}}>
+          <Typography variant="h4" sx={{mr:"auto"}}>{activity.title}</Typography>
+          
+          <Stack direction="row" spacing={1} sx={{height:"2.125rem"}}>
+            <Button variant="outlined" size="small" component={Link} href={`/${activity.isMission ? 'mission' : 'event'}/${eventId}/edit`}>Edit</Button>
+            <Button variant="outlined" size="small" onClick={() => setPromptingActivityState(true)}>{isActivityActive ? 'Complete' : 'Reactivate'}</Button>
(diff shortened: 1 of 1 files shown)


colincwilliams-kcsara:
[Separate status button from mission buttons (edit, complete, delete)](https://trello.com/c/vnssVRBy/20-separate-status-button-from-mission-buttons-edit-complete-delete)

src/app/(main)/EventPage.tsx
```diff
@@ -101,19 +101,25 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
     const isActivityActive = isActive(activity)
     body = (
       <Box>
-        <Typography variant="h4">{activity.title}</Typography>
+        <Stack direction="row" spacing={1} sx={{mt:2, mb:2}}>
+          <Typography variant="h4" sx={{mr:"auto"}}>{activity.title}</Typography>
+          
+          <Stack direction="row" spacing={1} sx={{height:"2.125rem"}}>
```
mcosand, line 107: We can use flex-box `align-items` to position the buttons without setting a specific size.
We can also use MUI breakpoints to change the stack direction from `column` to `row` for narrow screens.

```suggestion
        <Stack direction={{ xs: 'column', sm: 'row' }} spacing={1} alignItems="start" sx={{mt:2, mb:2}}>
          <Typography variant="h4" flexGrow={1}>{activity.title}</Typography>
          
          <Stack direction="row" spacing={1} alignItems="center">
```

image

image

mcosand (changes requested):
There will probably be times we need to use specific CSS, but would like to lean into MUI's layout and flex-box as much as we can.

src/app/(main)/EventPage.tsx
colincwilliams-kcsara: Very cool! Didn't realize these were all flex boxes. Happy to go with that.

mcosand (approved):

PR #6, opened 2023-06-04
Enforce user can only be signed into one mission at a time

mcosand:
A user should only be involved with one activity at a time. If they are already Signed Into another event, sign them out when signing in again. A user may keep Standby Status for multiple missions.


mcosand:
[Enforce user can only be signed into one mission at a time](https://trello.com/c/7PxnSIh9/24-enforce-user-can-only-be-signed-into-one-mission-at-a-time)

colincwilliams-kcsara (approved):

valonm (approved):

PR #7, opened 2023-06-05
Track miles

mcosand:
When signing out, the user has the option of entering total or additional miles:

image 
image

We don't currently have a report that includes miles, so the information goes into a little bit of a black hole

### DIFF
--- a/src/components/StatusUpdater/UpdateStatusForm.tsx
+++ b/src/components/StatusUpdater/UpdateStatusForm.tsx
@@ -1,0 +1,170 @@
+import { useState } from 'react';
+import { Box, DialogContentText, FormControl, FormHelperText, Radio, RadioGroup, TextField, Typography, Select, MenuItem, Stack } from '../Material';
+import { useAppDispatch } from '@respond/lib/client/store';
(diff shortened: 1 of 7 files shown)


mcosand:
[Miles](https://trello.com/c/AsFP3kkV/11-miles)

valonm (approved):

src/components/StatusUpdater/UpdateStatusForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+    errors: form.formState.errors,
+    doSubmit: form.handleSubmit(onSubmit),
+    context: {
+      newStatus,
+      activity,
+      participant,
+    },
+  };
+}
+type FormType = ReturnType<typeof useFormLogic>;
```
colincwilliams-kcsara, line 81: Maybe it's just me, but I found "FormType" to be a pretty confusing name. That made it sound like it's an enum or specialization of a generic form class (e.g. StatusForm, ContactForm, etc).

Best alternative that comes to mind is "FormLogic", tying into the name of the function. It's also fairly descriptive of what the type is.

PR #8, opened 2023-06-11
Numerous Improvements to the main UI

valonm:
This commit is implementing numerous UI enhancements:

- Refactored EventTile so that it can be used to implement the slight variations between all 3 sections.
  - It shows a standard set of information in all sections; additional elements can be slotted.
  - My Activity shows a reduced Mission Summary.
  - Missions show a full Mission Summary.
  - Trainings indicates "participants" instead of "responders".
- Deleted MyActivity and re-implemented with EventTile to reduce duplication of layout code and make the UI more consistent between the three sections.
- Added a StatusChip component to implement a responder status indicator.
- Added an OrganizationChip component to abstract the existing organization status indicators.

Note: Some of the screenshots show a "Navigate" button; this is commented for now to reserve the space.

image
image
image


src/app/(main)/StatusChip.tsx
```diff
@@ -0,0 +1,41 @@
+import { Chip } from '@mui/material';
+import { Circle } from '@mui/icons-material';
+import { ResponderStatus } from '@respond/types/activity';
+
+export const StatusChip = ({ status }: { status: ResponderStatus }) => {
+
+    switch (status) {
```
mcosand, line 7: Maybe a place for a couple lookup tables? Outside of the class:
```
const STATUS_COLORS = { ResponderStatus.SignedIn: 'success', ResponderStatus.SignedOut: 'error', ... };
const STATUS_TEXT = { ResponderStatus.SignedIn: 'Signed In', ... };
```
and then:
```
return (<Chip icon={<Circle color={STATUS_COLORS[status]} /> ... />);
```

mcosand (approved):

src/app/(main)/StatusChip.tsx
valonm: I had something similar originally, but TS was throwing an error. I couldn't find a resolution, so I just assigned them directly.

Any ideas on a better way to resolve this type mismatch?

image
image

valonm: To clarify -- even though this type checking error exists. It does not cause a runtime error.

mcosand: Ah, yes. We can explicitly define the types for these maps, instead of relying on TS inference. Pushed that as a commit.

PR #9, opened 2023-06-11
Fix Active Responder Count on EventTile

valonm:
The Active responder calculation used on `EventTile` is not calculating correctly. It looks like it always reflects the first status that was reported by each participant.

Looking further in the code, this is confirmed, the current `getActiveParticipants()` call is looking at the last element in the timeline array instead of the first; however the first is the most recent status.

I have updated the `getActiveParticipants()` calculation to evaluate the first status in the timeline array.

I have also added some `get<Status>Count()` methods for future use.

### DIFF
--- a/src/lib/client/store/activities.ts
+++ b/src/lib/client/store/activities.ts
@@ -1,5 +1,5 @@
 import { createSlice } from '@reduxjs/toolkit';
-import { Activity, ResponderStatus, ResponderUpdate } from '@respond/types/activity';
+import { Activity, Participant, ResponderStatus, ResponderUpdate } from '@respond/types/activity';
 import { RootState } from '.';
 import { ActivityState, ActivityActions, BasicReducers } from '@respond/lib/state';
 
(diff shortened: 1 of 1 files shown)


valonm:
[[BUG] "Active Responders" Calculation on EventTile](https://trello.com/c/dfCn8QqZ/27-bug-active-responders-calculation-on-eventtile)

src/lib/client/store/activities.ts
```diff
@@ -33,8 +33,24 @@ export function buildActivityTypeSelector(missions: boolean) {
...
+  return filterParticipantsByStatus(Object.values(activity.participants), [ResponderStatus.SignedIn]).length;
+}
+
+export function getStandbyCount(activity: Activity) {
+  return filterParticipantsByStatus(Object.values(activity.participants), [ResponderStatus.Standby]).length;
+}
+
+export function getSignedOutCount(activity: Activity) {
+  return filterParticipantsByStatus(Object.values(activity.participants), [ResponderStatus.SignedOut]).length;
+}
```
mcosand, line 50: I'm kind of thinking "you ain't gonna need it" here. I'd guess there's probably even odds that these implementations are going to be what we need next (vs mixing statuses, etc). Until then, we have to maintain methods with nothing pointing to them.

src/lib/client/store/activities.ts
```diff
@@ -33,8 +33,24 @@ export function buildActivityTypeSelector(missions: boolean) {
   return (state: RootState) => state.activities.list.filter(f => f.isMission === missions);
 }
 
-export function getActiveParticipants(a: Activity) {
-  return Object.values(a.participants).filter(p => (p.timeline.slice(-1)?.[0].status ?? ResponderStatus.Unavailable) & 0x01);
```
mcosand, line 37: oops. Thanks for the catch.

src/lib/client/store/activities.ts
valonm: Fair. I have some ideas for more EventTile data, but I will delete them for now. Easy enough to re-implement when we need them.

PR #13, opened 2023-06-15
Change : Limit Available Secondary Actions By Status

valonm:
I have swapped the handle menu item click handler to just invoke the onclick instead of requiring a second click.

### DIFF
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -30,7 +30,6 @@ function getRecommendedAction(current: ResponderStatus|undefined, startTime: num
 }
 
 function getCurrentOptions(current: ResponderStatus|undefined, startTime: number) {
-  console.log('!!!', current, startTime);
   if (isFuture(startTime)) {
     return current === ResponderStatus.Standby ? [options.standdown] : [options.standby];
   }
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -8,7 +8,6 @@ import { SplitButton } from '../SplitButton';
 import { useFormLogic, UpdateStatusForm } from './UpdateStatusForm';
 
 const options = {
-  standdown: { id: ResponderStatus.Cancel, text: 'Stand Down' },
   standby: { id: ResponderStatus.Standby, text: 'Stand By' },
   signin: { id: ResponderStatus.SignedIn, text: 'Sign In' },
   signout: { id: ResponderStatus.SignedOut, text: 'Sign Out' },
@@ -24,17 +23,17 @@ function getRecommendedAction(current: ResponderStatus|undefined, startTime: num
     return ResponderStatus.SignedOut;
   }
   if (isFuture(startTime)) {
-    return current === ResponderStatus.Standby ? ResponderStatus.Cancel : ResponderStatus.Standby;
+    return current === ResponderStatus.Standby ? ResponderStatus.SignedOut : ResponderStatus.Standby;
   }
   return ResponderStatus.SignedIn;
 }
 
 function getCurrentOptions(current: ResponderStatus|undefined, startTime: number) {
   if (isFuture(startTime)) {
-    return current === ResponderStatus.Standby ? [options.standdown] : [options.standby];
+    return current === ResponderStatus.Standby ? [options.signout] : [options.standby];
   }
   if (current === ResponderStatus.Standby) {
-    return [options.signin, options.standdown];
+    return [options.signin, options.signout];
   }
   if (current === ResponderStatus.SignedIn) {
     return [options.signout];
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -29,7 +29,6 @@ function getRecommendedAction(current: ResponderStatus|undefined, startTime: num
 }
 
 function getCurrentOptions(current: ResponderStatus|undefined, startTime: number) {
-  if (current === undefined) { return options; }
   if (isFuture(startTime)) {
     if (current === ResponderStatus.Standby) {
       return options.filter(option => option.id === ResponderStatus.SignedOut);
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -8,22 +8,38 @@ import { SplitButton } from '../SplitButton';
 import { useFormLogic, UpdateStatusForm } from './UpdateStatusForm';
 
 const options = [
-  { id: ResponderStatus.Unavailable, text: 'Not Available' },
   { id: ResponderStatus.Standby, text: 'Stand By' },
   { id: ResponderStatus.SignedIn, text: 'Sign In' },
   { id: ResponderStatus.SignedOut, text: 'Sign Out' },
 ]
 const optionTexts = options.reduce((accum, cur) => ({ ...accum, [cur.id]: cur.text }), {} as Record<string, string>);
 
+function isFuture(startTime: number) {
+  return (startTime - 60 * 60 * 1000) > new Date().getTime();
+};
+
 function getRecommendedAction(current: ResponderStatus|undefined, startTime: number): ResponderStatus {
-  const now = new Date().getTime();
   if (current === ResponderStatus.SignedIn) {
     return ResponderStatus.SignedOut;
   }
-  if (startTime - 60 * 60 * 1000 > now) {
-    return ResponderStatus.Standby;
+  if (isFuture(startTime)) {
+    return current === ResponderStatus.Standby ? ResponderStatus.SignedOut : ResponderStatus.Standby;
   }
   return ResponderStatus.SignedIn;
+}
+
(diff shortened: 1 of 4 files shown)


valonm:
I made additional updates so that the drop down only shows statuses that the user should be able to move to.

src/components/StatusUpdater/index.tsx
```diff
@@ -8,24 +8,40 @@ import { SplitButton } from '../SplitButton';
...
 
 const options = [
-  { id: ResponderStatus.Unavailable, text: 'Not Available' },
   { id: ResponderStatus.Standby, text: 'Stand By' },
   { id: ResponderStatus.SignedIn, text: 'Sign In' },
   { id: ResponderStatus.SignedOut, text: 'Sign Out' },
 ]
 const optionTexts = options.reduce((accum, cur) => ({ ...accum, [cur.id]: cur.text }), {} as Record<string, string>);
 
+function isFuture(startTime: number) {
```
colincwilliams-kcsara, line 17: nit: parameter name here can be more generic - just "time". StartTime just happens to be what's passed in.

src/components/StatusUpdater/index.tsx
```diff
@@ -8,24 +8,40 @@ import { SplitButton } from '../SplitButton';
...
 
+function getCurrentOptions(current: ResponderStatus|undefined, startTime: number) {
+  if (current === undefined) { return options; }
+  if (isFuture(startTime)) {
+    if (current === ResponderStatus.Standby) {
+      return options.filter(option => option.id === ResponderStatus.SignedOut);
+    }
+    return options.filter(option => option.id === ResponderStatus.Standby);
+  }
+  if (current === undefined || current === ResponderStatus.Unavailable) {
```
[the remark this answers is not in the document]
valonm: Hmm... I didn't realize that guard clause was at the top, that actually needs to be below the future use case and should never return all options. I removed the unavailable status as an option here, but it's still a responder status and I was seeing records come through in that state when the user hasn't interacted with that mission yet. I think both cases need to be handled, here, but I need to remove the guard clause above. They have to be handled as a separate use case because sign out shouldn't be an option to a mission that you haven't signed into.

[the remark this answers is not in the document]
colincwilliams-kcsara: Ahh ok I'm on board now. Agreed.

colincwilliams-kcsara (approved):

valonm:
@mcosand-caltopo Agreed, we should make that a single motion -- it should result in both a sign out and a sign in. However, I'm not sure how the data model works in this regard. The participant record is tied to an org id. If they sign in as a different unit are they a different entity?

image

I think this is an existing problem that is outside the scope of this change. If anything, this change would enforce that they have to sign out before they can sign back in, instead of allowing signed in => signed in. Something to think about as a further enhancement, though.

valonm:
The most recent commit implements ResponderStatus.Cancel, which is assigned by the "Stand Down" action in the status updater.

If you are standing by for a mission, you can either stand down, or sign in. standing down moves you to cancelled status. this preserves sign in -> sign out relationship.

image

In a separate commit I am going to update the status indicator to show your last signed in, standby or signed out state so that if you sign out of a mission, then go on standby, then cancel, the sign out status indicator will return. at present, the Cancel status does not show an indicator when it is your most recent status update.

src/types/activity.ts
```diff
@@ -14,6 +14,7 @@ const pickSafely = <ObjectType>(keys: readonly `${string & keyof ObjectType}`[])
 
 export enum ResponderStatus {
   Unavailable = 0,
+  Cancel = 1,
```
mcosand, line 17: I think this commit has some big implications we need to think through. I don't understand what problem we're trying to solve.

The original intent was that `ResponderStatus` is a state, but this change seems to focus on `Cancel` as a state transition (from StandBy to Cancel). If we're going to track each state transition, I think this enum (and supporting code) will get quite messy.

If we say "SignedOut" is "no longer involved with the mission", what's wrong with moving from Standby ("I am interested / planning on going on this mission") to SignedOut?

If there's a concern that we need an explicit transition from StandBy to SignedOut, I'd be less worried about code that changes the text of the "Sign Out" action based on the participant's current state, but I feel like that's unnecessary complexity.

mcosand (changes requested):

mcosand:
> @mcosand-caltopo Agreed, we should make that a single motion -- it should result in both a sign out and a sign in. However, I'm not sure how the data model works in this regard. The participant record is tied to an org id. If they sign in as a different unit are they a different entity?
> 
> image
> 
> I think this is an existing problem that is outside the scope of this change. If anything, this change would enforce that they have to sign out before they can sign back in, instead of allowing signed in => signed in. Something to think about as a further enhancement, though.

The organizationId is tracked in the timeline entries. I'd have to do some reading to figure out why/how it got promoted to the participant, but there's a good chance it doesn't need to be. I'll take a look at this.

src/types/activity.ts
valonm: I'm thinking of it more as "not responding"; as a state, not the state transition. That's actually why I didn't call it "Stand Down". I do think this is different than Signed Out, in terms of the timeline. "Cancel" was just what came to mind, but since you've prompted more thought, "NotResponding" could work, too.

I also feel like its a different state than "Unavailable", but, I'm not totally sure what the intent of unavailable was.

**This is how I'm thinking about each of the statuses:**
Not Responding (Cancel) means "I'm not currently responding to the mission"
Standby means "I am not currently responding to the mission, but I will be"
Signed In means "I'm currently responding to the mission"
Signed Out means "I was responding to the mission, but I'm not now"

The extra state of Not Responding provides a means to cancel a standby without adding an orphaned sign out to the history -- there will always be a corresponding sign in.

This has implications on the UI, and simplifies the necessary logic for things like the status indicator. A mission that you were on stand by but stood down for should not show signed out. Where as a mission that you signed into and out of, should.

Also, I think it makes a better user experience. Its more clear that I was never signed in; and I don't need the sign out dialog if I never responded.

valonm:
The scope of this commit creeped a ways. I'm thinking I'll pull out the changes related to the new status, and defer until further discussion. The options when you are in standby will be sign in and sign out.

@mcosand-caltopo did you have any other concerns that weren't related to that new status?

mcosand:
> The scope of this commit creeped a ways. I'm thinking I'll pull out the changes related to the new status, and defer until further discussion. The options when you are in standby will be sign in and sign out.
> 
> @mcosand-caltopo did you have any other concerns that weren't related to that new status?

I'll need to look at this again when I get more time. I'm still trying to convince myself that trimming the options is appropriate complexity for this point in the project, but I'm coming around. I do think I've come around to altering Material's default split button to click-on-menu-select.

valonm:
FWIW, MUI's Button Group docs say that the Split Button can be used to either change the primary button, or to immediately trigger a related action. The template implements the former; but I don't think my changes are necessarily breaking the design pattern.

https://mui.com/material-ui/react-button-group/#split-button
> ButtonGroup can also be used to create a split button. The dropdown can change the button action (as in this example) or be used to immediately trigger a related action.

Personally, I don't like the default functionality. I just want it to do the thing, the first time I ask. It's an extra click and I find the fact that the button value changes a little bit disjointed from a UX perspective.

valonm:
I've separated the scope of the immediate secondary menu action to a separate commit. This commit is now only concerned with the options that are available in the dropdown.

src/components/StatusUpdater/index.tsx
```diff
@@ -7,25 +7,43 @@ import { UserInfo } from '@respond/types/userInfo';
...
+  if (current === ResponderStatus.Standby) {
+    return [options.signin, options.signout];
+  }
+  if (current === ResponderStatus.SignedIn) {
+    return [options.signout];
+  }
+  if (current === ResponderStatus.SignedOut) {
+    return [options.signin, options.standby];
+  }
+  return [options.signin, options.standby];
```
davidgut76, line 44: Could the non-future start time transitions be described with a const map of some kind? Another more readable option may be to use a switch statement.

src/components/StatusUpdater/index.tsx
```diff
@@ -7,25 +7,43 @@ import { UserInfo } from '@respond/types/userInfo';
...
+const options = {
+  standby: { id: ResponderStatus.Standby, text: 'Stand By' },
+  signin: { id: ResponderStatus.SignedIn, text: 'Sign In' },
+  signout: { id: ResponderStatus.SignedOut, text: 'Sign Out' },
+}
+const optionTexts = Object.values(options).reduce((accum, cur) => ({ ...accum, [cur.id]: cur.text }), {} as Record<string, string>);
+
+function isFuture(time: number) {
+  return (time - 60 * 60 * 1000) > new Date().getTime();
+};
```
davidgut76, line 19: Maybe put 60 * 60 * 1000 into a const so it's self-documenting. I'm guessing it's millisecondsPerHour?

davidgut76 (commented):
A couple of comments, but I'm new here, so take them with a grain of salt.

PR #16, opened 2023-06-19
Change : SplitButton to Immediately Trigger Secondary Action

valonm:
- Change SplitButton to immediately trigger the Secondary Action.
- Remove Primary Action from the menu.

### DIFF
--- a/src/components/SplitButton.tsx
+++ b/src/components/SplitButton.tsx
@@ -33,7 +33,7 @@ export function SplitButton<K extends string|number|EnumMember, T extends IdOpti
     event: React.MouseEvent<HTMLLIElement, MouseEvent>,
     index: number,
   ) => {
-    setSelectedIndex(index);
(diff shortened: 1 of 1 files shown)


valonm:
Copied from the dropdowns commit discussion:

MUI's Button Group docs say that the Split Button can be used to either change the primary button, or to immediately trigger a related action. The template implements the former; but I don't think my changes are necessarily breaking the design pattern.

https://mui.com/material-ui/react-button-group/#split-button
> ButtonGroup can also be used to create a split button. The dropdown can change the button action (as in this example) or be used to immediately trigger a related action.

Personally, I don't like the default functionality. I just want it to do the thing, the first time I ask. It's an extra click and I find the fact that the button value changes a little bit disjointed from a UX perspective.

mcosand-caltopo (approved):

colincwilliams-kcsara (approved):

PR #21, opened 2023-06-21
Display an error if the user's email address is not found in D4H

colincwilliams-kcsara:
Added support for displaying error messages during login.

Primary error to handle is when the user's email cannot be found in D4H (yet they were able to authenticate with Google). It is clearly actionable and has been hit before. Handled other errors gracefully.

Validated that the error is gone when returning to the login page after a successful sign-in and signing back out.

image

### DIFF
--- a/src/app/LoginPanel.tsx
+++ b/src/app/LoginPanel.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { Box, Button, Stack, Typography } from '@respond/components/Material';
+import { Box, Button, Stack } from '@respond/components/Material';
 import { CredentialResponse, GoogleLogin } from '@react-oauth/google';
 import Api from '@respond/lib/api';
 import { useAppDispatch, useAppSelector } from '@respond/lib/client/store';
@@ -29,8 +29,7 @@ export default function LoginPanel() {
   }
 
   async function finishLogin(token: string) {
-    const res1 = await Api.post<any>('/api/auth/google', { token });
-    const res = res1 as AuthResponse;
+    const res = await Api.post<any>('/api/auth/google', { token }) as AuthResponse;
     console.log('login response', res);
 
     if (res.userInfo) {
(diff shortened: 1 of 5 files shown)


colincwilliams-kcsara:
[Display an error if the user's email address is not found in D4H](https://trello.com/c/kE5kN5Rg/31-display-an-error-if-the-users-email-address-is-not-found-in-d4h)

mcosand (approved):

src/app/LoginPanel.tsx
```diff
@@ -25,22 +30,46 @@ export default function LoginPanel() {
...
+
+    if (res.userInfo) {
+      localStorage.userAuth = JSON.stringify(res.userInfo);
+      dispatch(AuthActions.set({ userInfo: res.userInfo }));
+      dispatch(OrgActions.set({ mine: res.organization }));
+    } else {
+      switch (res.error) {
+        case AuthError.USER_NOT_KNOWN:
+          setError("Could not find email address in D4H");
+          setErrorDetails("Please contact your unit's database manager to have your email address added to your D4H profile, or log in with an email address in your profile.");
```
mcosand, line 43: Nit - it's not a given that the organization is using D4H as their membership store.

colincwilliams-kcsara: Hm. Good point. I don't know if there's a way to word this without using "D4H" (or whatever the provider is) while keeping it directly useful to users.

Can pass the member provider name through site config, but then that violates the principle of keeping org data off the client when the user is unauthenticated.

colincwilliams-kcsara: Going to get this in as-is for the demo. Given our discussions in Slack, I'm not concerned about leaking that D4H is in use since the error message will only appear after oauth succeeds.

I'll fast follow with a change to send down the org info for these error messages and customize the error text based on that info.

PR #22, opened 2023-06-23
Update readme

davidgut76:
Update README 
- reorder sections for a more logical flow
- add details in places needing more help
- mark a few things as optional when possible

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,15 @@
 ## Configure Development Environment
 
 ### Install dev/build tools
+- [Node.js] (https://nodejs.org/en) 16.x or higher
 - [Yarn](https://yarnpkg.com)
-- Node.js 16.x or higher (installed with yarn)
 
 ### Create dev database
 Development is supported on either a local or remote MongoDB server.
 - For a local installation, it's recommended to install the [Community Server](https://www.mongodb.com/try/download/community) and [Compass UI](https://www.mongodb.com/try/download/compass).
 - For a remote DB, use [Mongo Atlas](https://www.mongodb.com/atlas/database). You can also use [Compass UI](https://www.mongodb.com/try/download/compass), but isn't strictly necessary. 
 
-In your MongoDB, add a database named `respond-dev`. Then also add a user for that database. In Mongo Atlas this is done by clicking on "Database Access" from the left side bar navigation. Then click the "Add new database user" and give them read/write permission. This username/password is what you'll use in your connection string below. 
+In your MongoDB, add a database named `respond-dev`. If you're using Mongo Atlas, also add a user for that database. Click on "Database Access" from the left side bar navigation, then click the "Add new database user" and give them read/write permission. This username/password is what you'll use in your connection string below. 
 
 #### Add Seed Data to MongoDB
 The app is multi-tenant, and matches hostnames to organizations stored in the database. To be able to use the application, at least one organization must be created in the database. Below are two documents that should allow you to get up and running.
@@ -24,7 +24,7 @@ Try to get a key yourself:
 
 If this fails, contact your database admin to ask for a key. 
 
-To fully test multi-tenancy you'll also need to be able to refer to the site by multiple host names. It may be possible to do this with a combination of `localhost` and your computer's hostname. You may also want/need to add aliases to your local machine in your `/etc/hosts` file or router DNS configuration. These hostnames will be used in the seed data. However this is optional for an initial setup. 
+To fully test multi-tenancy you'll also need to be able to refer to the site by multiple host names. It may be possible to do this with a combination of `localhost` and your computer's hostname. You may also want/need to add aliases to your local machine in your hosts file (`/etc/hosts` on Linux or `C:\Windows\System32\drivers\etc\hosts` on Windows) or router DNS configuration. These hostnames will be used in the seed data. However this is optional for an initial setup. 
 
 In your MongoDB database, make sure there is a collection called `organizations` and add the following two documents. Note the two placeholders needing replacement. The first is usually "localhost", and the second is your API key:
 ```json
--- a/README.md
+++ b/README.md
@@ -7,21 +7,22 @@
 ### Create dev database
 Development is supported on either a local or remote MongoDB server.
 - For a local installation, it's recommended to install the [Community Server](https://www.mongodb.com/try/download/community) and [Compass UI](https://www.mongodb.com/try/download/compass).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,69 +1,30 @@
...
-
-## Seed Data
+#### Add Seed Data to MongoDB
 The app is multi-tenant, and matches hostnames to organizations stored in the database. To be able to use the application, at least one organization must be created in the database. Below are two documents that should allow you to get up and running.
 
-The application does have integration with the organization's membership database (currently D4H). In order to set this up you will need an API key from your member database admin.
+The application also has integration with D4H, so you will need an API key, either generated yourself or from your member database admin.
+**Note that a D4H API key is equivalent to your password: it provides full access to personal information for every KCSARA member - even the bits that are marked ‘private’. Make sure you don't make this key public (e.g. by checking it in).**
+Try to get a key yourself:
+- Go to https://kcsara.d4h.org/members/api_settings
```
mcosand, line 20: Please don't put this KCSARA specific URL In the public repo

mcosand (changes requested):
It seems the hostname edits are adding data that's not necessary. I'd like to figure this out before it becomes canon in the readme. I don't think the following hosts do anything, and can be removed: `http://localhost`, `https://localhost:3000`, `http://localhost:9001`

README.md
davidgut76: Switched this to use navigation instructions instead.

davidgut76: I wasn't able to authenticate without both http://localhost:3000 and http://locahost as allowed domains. Leaving out http://localhost made the Google Auth window show up entirely blank every time I tried it, which makes me think there's something else weird going on. 
I haven't gotten this working with the proxy, so I don't know the proxy needs both https URLs added.

README.md
```diff
@@ -1,69 +1,32 @@
 ## Configure Development Environment
 
 ### Install dev/build tools
-- Node.js 16.x or higher
 - [Yarn](https://yarnpkg.com)
+- Node.js 16.x or higher (installed with yarn)
```
colincwilliams-kcsara, line 5: I'm confused by "installed with Yarn". Both install paths I'm seeing for Yarn require having Node installed first.

[Classic instructions](https://classic.yarnpkg.com/en/docs/install#windows-stable) are to run `npm install --global yarn`. Quoting that page:

> It is recommended to install Yarn through the npm package manager, which comes bundled with Node.js when you install it on your system.

[New instructions](https://yarnpkg.com/getting-started/install) use corepack: `corepack enable; corepack prepare yarn@stable --activate`. Corepack also requires Node to be installed first:

> Corepack is included by default with all Node.js installs

Is there a documented/preferred way of installing Yarn first and then using Yarn to install Node.js?

README.md
```diff
@@ -1,69 +1,32 @@
...
-#### MongoDB Atlas service:
-Add the connection string for your database to `.env.local`:
-```
-MONGODB_URI="mongodb+srv://database:<your-connection-string>.mongodb.net/respond-dev?retryWrites=true&w=majority"
-```
+Development is supported on either a local or remote MongoDB server.
+- For a local installation, it's recommended to install the [Community Server](https://www.mongodb.com/try/download/community) and [Compass UI](https://www.mongodb.com/try/download/compass).
+- For a remote DB, use [Mongo Atlas](https://www.mongodb.com/atlas/database). You can also use [Compass UI](https://www.mongodb.com/try/download/compass), but isn't strictly necessary. 
 
+In your MongoDB, add a database named `respond-dev`. Then also add a user for that database. In Mongo Atlas this is done by clicking on "Database Access" from the left side bar navigation. Then click the "Add new database user" and give them read/write permission. This username/password is what you'll use in your connection string below.
```
colincwilliams-kcsara, line 12: nit: Adding a user is only necessary for a remote database.

README.md
```diff
@@ -1,69 +1,32 @@
...
+Try to get a key yourself:
+- Login to D4H
+- Click on the small head icon in the upper right and go to "My Settings"
+- Click on API access keys
+- Give your key a name and click "Generate API Access Key". You'll use this key in the next section. 
+
+If this fails, contact your database admin to ask for a key. 
 
-To fully test multi-tenancy you'll also need to be able to refer to the site by multiple host namess. It may be possible to do this with a combination of `localhost` and your computer's hostname. You may also want/need to add aliases to your local machine in your `/etc/hosts` file or router DNS configuration. These hostnames will be used in the seed data.
+To fully test multi-tenancy you'll also need to be able to refer to the site by multiple host names. It may be possible to do this with a combination of `localhost` and your computer's hostname. You may also want/need to add aliases to your local machine in your `/etc/hosts` file or router DNS configuration. These hostnames will be used in the seed data. However this is optional for an initial setup.
```
colincwilliams-kcsara, line 27: Would it make sense to switch
> `/etc/hosts` file

To just

> hosts file

I propose this because the Windows hosts file isn't located there, and the goal is that development should work on all platforms. It seems overkill to me to provide instructions for editing the hosts file on both *nix and Windows, as it's easy enough to look up.

(Though I'm not strongly against providing instructions for both platforms.)

README.md
colincwilliams-kcsara, line 12: Worth noting that these steps are only required if you either didn't use the database creation wizard. If you used the wizard it forces you to create a user and provide a password, in which case you should use those credentials.

If you don't remember the password you used for that user, you can go to the same place in settings and change that user's password to something you remember.

colincwilliams-kcsara (changes requested):
Fantastic improvements! Just want to clarify some of the changes before they go in.

README.md
davidgut76: Updated.

README.md
davidgut76: I left this because the path is nearly correct, even for Windows. Switched this to include the path for both Linux and Windows.

README.md
davidgut76: Is there a database wizard for Atlas? Perhaps I just skipped that...

README.md
colincwilliams-kcsara: Perfect. Thanks!

README.md
colincwilliams-kcsara: There is. But I only encountered it when I first setup an account (when we were debugging your setup). Couldn't figure out how to get back to it. It can indeed be skipped.

davidgut76: Added a note about this too. Not sure how much I can say about it since neither of us has tried it.

PR #23, opened 2023-06-27
layout-components : Implement OutputForm and OutputField

valonm:
Implements a pair of generic components for rendering labelled values; object fields such as Mission attributes. These will help make better use of the available space and take the boilerplate out of rendering/formatting this type of data.

Example Markup:
~~~~
<OutputForm>
  <OutputField label="Location" value={a.location.title} />
  <OutputField label="State #" value={a.idNumber} />
  <OutputField label="Active Responders" value={getActiveParticipants(a).length.toString()} />
  <OutputField label="Status" value={a.status} />
</OutputForm>
~~~~

This commit does not include changes to EventTile or the Event page; screenshots are for demonstration only.

image
image

### DIFF
--- a/src/app/(main)/EventTile.tsx
+++ b/src/app/(main)/EventTile.tsx
@@ -10,20 +10,19 @@ export const EventTile = ({ activity, status, children }: { activity: Activity, 
 
     return (
       <Card>
-        <Box sx={{ p: 1, display: "flex", flexDirection: "row" }}>
+        <Box paddingX={1} paddingTop={1} sx={{ display: "flex", flexDirection: "row" }} alignItems="center">
           <Box sx={{ flexGrow: 1 }}>
             <Link href={getActivityPath(activity)} color="textPrimary" underline="hover">
               <Typography sx={{ fontWeight: "bold" }} variant="h6">
                 {activity.title}
               </Typography>
             </Link>
-            <Box>Location: {activity.location.title}</Box>
-            <Box>{children}</Box>
           </Box>
-          <Box sx={{ alignSelf: "flex-start" }}>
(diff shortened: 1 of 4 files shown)


colincwilliams-kcsara:
Love this!

I know you explicitly stated that your PR doesn't have changes outside of adding these components, but why not? Would be nice to see them in use in the project at least once, even if you don't switch everywhere all at once (to scope the PR).

valonm:
> Love this!
> 
> I know you explicitly stated that your PR doesn't have changes outside of adding these components, but why not? Would be nice to see them in use in the project at least once, even if you don't switch everywhere all at once (to scope the PR).

Just trying to avoid scope creep. These components are generic. There are a few places I want to implement them.

colincwilliams-kcsara:
Ok, sounds good. As long as there's explicitly planned places for them to be used then that works. I don't want dead code being checked-in. :)

colincwilliams-kcsara (approved):

valonm:
@colincwilliams-kcsara :-D you approved it while I was busy implementing on the event page.

The Event Page will now implement the existing fields as part of this commit.

image

mcosand:
(this was sitting as a draft. Don't know why it wasn't sent. Sending anyways for thoughts about how we add code)

I think it would be fine to include these as other work. It's kind of hard to evaluate some of the edges when we don't have more context (should OutputField pass `sx` props so we can make fields responsive and show/hide based on screen size, etc).

I'm okay with merging this request, or keeping this PR open and appending feature changes to it, or putting this change on a feature branch where it can collect smaller PRs until ready for `main`.

PR #26, opened 2023-06-28
responder-status : Refactor ResponderStatus to Bit Flags

valonm:
This commit refactors ResponderStatus as a Bit Flag Enum.

Subject to change, I expect there will be continued discussion around this. Just getting a branch started that we can iterate on.

I'm also not sure what the best way to document the bitwise flagging is. I would think there is probably a best practice around this, like some sort of pattern to make the code self documenting, or a commenting strategy, but I haven't found it yet.

https://docs.google.com/spreadsheets/d/1QBt3RkDtGucOKESqMf_QjtJSwyThdayMig1qjIQYURQ/edit#gid=0

### DIFF
--- a/src/app/(main)/EventPage.tsx
+++ b/src/app/(main)/EventPage.tsx
@@ -26,7 +26,7 @@ const Roster = ({participants, orgs, startTime}: {participants: Record<string, P
   };
   
 
-  const rows: GridRowsProp = Object.values(participants).filter(f => f.timeline[0].status !== ResponderStatus.Unavailable).map(f => ({
+  const rows: GridRowsProp = Object.values(participants).filter(f => f.timeline[0].status !== ResponderStatus.NotResponding).map(f => ({
     ...f,
     orgName: orgs[f.organizationId]?.rosterName ?? orgs[f.organizationId]?.title,
     status: f.timeline[0].status,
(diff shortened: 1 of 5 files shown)


colincwilliams-kcsara:
What's the motivation behind refactoring to bit flags and changing the values of existing statuses? It's a breaking change with regards to existing data in the database, so I'm hesitant to do it without a strong reason to do so.

Additionally, as a result of that, this change requires a migration plan. That could be implemented in code, so it's an automatic migration, or a planned manual migration of the data in the database. I lean towards the former to reduce the chance of human error and race conditions, but both paths have their merits.

To be clear: I'm all for updating the statuses to include more of them and make them fit our goals. What I'm concerned about is changing the values of existing statuses rather than just appending to the enum numerically.

valonm:
There's a whole tread on this in slack that I don't want to move here. I tagged you on it.

colincwilliams-kcsara:
Got it. Thanks! Will move the discussion there for now. Biggest heads up is that this will need a migration plan before it can go in (once we're settled on the values).

valonm:
>This will need a migration plan
Meh, we haven't really started using the tool yet; save for a couple of practice missions. having the old values in there wont break anything. We could run a script to update them if we want to keep that for some reason; OR we could just delete all the activities in the database after deploying this. Leaning toward the latter, personally.

PR #29, opened 2023-07-01
Event page

valonm:
Updated Event Page Layout and Styling
- Added Margin at the bottom, when the page is longer than the viewport.
- Implemented the previously deployed OutputForm elements
- Added OutputTime
- Moved the content into a Paper component because I like how it looks.

image


src/components/OutputTime.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const interval = setInterval(() => setNowTime(new Date().getTime()), 5000);
+    return () => {
+      clearInterval(interval);
+    }
+  });
+
+  return (
+    <Box sx={{ display: "flex", flexDirection: "row", alignItems: "center", justifyContent: "space-between", borderColor: 'grey.200' }} borderBottom={1} >
+      <Typography variant="body2" sx={{ color: 'text.secondary', lineHeight: "2em" }}>{label}:</Typography>
+      {time && <RelativeTimeText time={time ?? 0} baseTime={nowTime} defaultToTime={!relative} />}
```
colincwilliams-kcsara, line 19: I feel like it should be possible to pass the RelativeTimeText object into the value of OutputField somehow so that the rest of the presentation isn't duplicated. Not quite sure how at this moment, but should be possible.

colincwilliams-kcsara (changes requested):

valonm:
@colincwilliams-kcsara I refactored the form components so that they compose a base element for each of the discrete types. That way, only the discrete types are only implementing the part that is different and they all share the same presentation element.

colincwilliams-kcsara (approved):

PR #30, opened 2023-07-01
Customized Error Messages

colincwilliams-kcsara:
Adds support for error messages that are customized in two ways:

1. Adds support for a new, optional "supportEmail" field for organizations. This is used in error messaging to direct users how to get support.
2. Using the member provider's name in error messaging.

As a part of supporting the member provider's name, this PR enforces type safety for member provider types rather than using a bare string.

image
image
image

### DIFF
--- a/src/app/LoginPanel.tsx
+++ b/src/app/LoginPanel.tsx
@@ -39,28 +39,30 @@ export default function LoginPanel() {
       dispatch(OrgActions.set({ mine: res.organization }));
     } else {
       const organization = res.organization;
-      let supportEmail: string | undefined = undefined;
+      let supportContact: string | undefined = undefined;
       let memberProviderName = undefined;
 
       if (organization) {
-        supportEmail = organization.supportEmail;
+        supportContact = organization.supportEmail;
         memberProviderName = MemberProviderName[organization.memberProvider];
       }
+
+      supportContact ??= "your unit's operations leader";
 
       switch (res.error) {
         case AuthError.USER_NOT_KNOWN:
           if (memberProviderName) {
             setError(`Could not find your email address in ${memberProviderName}`);
-            setErrorDetails(`Please log in with an email address in your ${memberProviderName} profile. If your email address is not in your profile, contact ${supportEmail ?? "your unit's operations leader"} for support.`);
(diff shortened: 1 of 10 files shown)


colincwilliams-kcsara:
[Display an error if the user's email address is not found in D4H](https://trello.com/c/kE5kN5Rg/31-display-an-error-if-the-users-email-address-is-not-found-in-d4h)

src/app/LoginPanel.tsx
```diff
@@ -39,28 +39,30 @@ export default function LoginPanel() {
...
+      let supportContact: string | undefined = undefined;
       let memberProviderName = undefined;
 
       if (organization) {
-        supportEmail = organization.supportEmail;
+        supportContact = organization.supportEmail;
         memberProviderName = MemberProviderName[organization.memberProvider];
       }
 
+      supportContact ??= "your unit's operations leader";
```
valonm, line 50: Operations Leader seems too specific to our org structure. Maybe something more generic like system administrator?

colincwilliams-kcsara: Operations Leader was suggested somewhere in Slack before. The problem with System Administrator is I don't think anyone would know who that is unless they're already involved in the tech side of things.

When you say "our org structure", who are you referring to? I know that applies to 4x4 and ESAR. Doesn't it apply to all units in KCSARA?

valonm: I just don't think an operations leader is the right person to contact, and if we go outside of King county other organizations may not have that title or role

valonm: WASART is a possible example that I could see us supporting in the nearer term. They do not have "OLs".

colincwilliams-kcsara: Ahh and of course in my head I'm thinking "leader of operations" administratively...i.e. VPO or Operations Chair. But there's not really a standard term for that either.

I guess I can use System Administrator and just trust the user to escalate up their chain of command appropriately. Will make that change.

colincwilliams-kcsara: Done.

valonm (approved):

PR #33, opened 2023-07-03
Support per-org favicon

colincwilliams-kcsara:
Adding support for an organization to define a favicon and home screen icon (for when the site is pinned to a home screen on mobile). There are no filesystem changes (e.g. adding images) as the intent is organizations will set the icon to match their publicly facing website (or an image file hosted there).

While the optimal home screen icon solution would require responding with a custom site manifest for Android, by my research it will happily fall back to apple-touch-icon if the site manifest isn't present. I'll come back and add support for the site manifest if this doesn't work well.

### DIFF
--- a/src/types/data/organizationDoc.ts
+++ b/src/types/data/organizationDoc.ts
@@ -23,6 +23,8 @@ export interface OrganizationDoc {
   brand: {
     primary: string;
     primaryDark?: string;
+    favicon?: string;
+    homeScreenIcon?: string;
   };
   memberProvider: D4HConfig;
   canCreateEvents: boolean;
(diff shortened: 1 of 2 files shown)


colincwilliams-kcsara:
[Set app icon/favicon per organization for when the site is added to a phone home screen](https://trello.com/c/PCpu5Qbc/30-set-app-icon-favicon-per-organization-for-when-the-site-is-added-to-a-phone-home-screen)

src/types/data/organizationDoc.ts
```diff
@@ -23,6 +23,8 @@ export interface OrganizationDoc {
   brand: {
     primary: string;
     primaryDark?: string;
+    favicon?: string;
```
valonm, line 26: Suggestion: It's a bit unclear what this field needs; maybe something like `faviconUrl`, `faviconSrc` or `faviconHref` would make this more self-documenting.

colincwilliams-kcsara: Good call! Done.

valonm (approved):

PR #34, opened 2023-07-03
Consistently display 24 hour time

colincwilliams-kcsara:
We weren't displaying 24 hour time for the start and end times for activities. This is because they were using date-fns formatRelative method, which displays 12 hour time by default without an obvious way to make it 24 hour time.

Figured out how to make formatRelative display 24 hour time. It does mean that formatRelative is now specific to enUS; however, I feel that's appropriate for the audience of this app for the foreseeable future.

Additionally, removed the colon from time display to align with roster time formatting and for cleanliness.

image


colincwilliams-kcsara:
[Time Format 24 Hours](https://trello.com/c/vK3zh2kj/39-time-format-24-hours)

valonm (approved):

davidgut76:
Thanks for fixing this, it's actually one of the first things I noticed (but wasn't sure how to fix).

I added a colon into the roster time to be consistent... which was totally wrong in retrospect. Can you remove that colon as part of this? It's at EventPage.tsx:51

colincwilliams-kcsara:
Done!

PR #35, opened 2023-07-03
Change : Make OutputTextArea collapsible

valonm:
Added a `maxHeight` prop on OutputTextArea
- When `maxHeight` is `undefined`, renders the full text.
- When `maxHeight` is set, the max height of the field is limited
  - A show more/show less button appears. Clicking toggles the full text.
  - Clicking anywhere on the body of the field also toggles the full text.
- I had to revert the line height of empty lines to 1 to avoid rendering half-lines.

Collapsed
image

Expanded:
image

### DIFF
--- a/src/app/(main)/EventPage.tsx
+++ b/src/app/(main)/EventPage.tsx
@@ -115,7 +115,7 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
             <OutputTime label="End Time" time={activity.endTime}></OutputTime>
           </Box>
         </OutputForm>
-        <OutputTextArea label="Description" value={activity.description} maxHeight={3}></OutputTextArea>
+        <OutputTextArea label="Description" value={activity.description} rows={3}></OutputTextArea>
 
         <Box sx={{ my:2 }}>
           {isActivityActive && <StatusUpdater activity={activity} current={myParticipation?.timeline[0].status} />}
--- a/src/app/(main)/EventPage.tsx
+++ b/src/app/(main)/EventPage.tsx
@@ -115,7 +115,7 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
             <OutputTime label="End Time" time={activity.endTime}></OutputTime>
           </Box>
         </OutputForm>
-        <OutputTextArea label="Description" value={activity.description}></OutputTextArea>
+        <OutputTextArea label="Description" value={activity.description} maxHeight={3}></OutputTextArea>
 
         <Box sx={{ my:2 }}>
           {isActivityActive && <StatusUpdater activity={activity} current={myParticipation?.timeline[0].status} />}
--- a/src/components/OutputForm.tsx
+++ b/src/components/OutputForm.tsx
@@ -1,4 +1,4 @@
-import { Box, Typography, Grid, Button } from '@mui/material';
+import { Box, Typography, Grid } from '@mui/material';
 import React, { ReactNode } from 'react';
 import Link from 'next/link';
 import { useEffect, useState } from 'react';
--- a/src/components/OutputForm.tsx
+++ b/src/components/OutputForm.tsx
@@ -1,9 +1,10 @@
-import { Grid } from '@mui/material';
+import { Box, Typography, Grid, Button } from '@mui/material';
 import React, { ReactNode } from 'react';
-import { Box, Typography } from '@mui/material';
 import Link from 'next/link';
 import { useEffect, useState } from 'react';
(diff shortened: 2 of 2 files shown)


src/app/(main)/EventPage.tsx
```diff
@@ -115,7 +115,7 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
             <OutputTime label="End Time" time={activity.endTime}></OutputTime>
           </Box>
         </OutputForm>
-        <OutputTextArea label="Description" value={activity.description}></OutputTextArea>
+        <OutputTextArea label="Description" value={activity.description} maxHeight={3}></OutputTextArea>
```
colincwilliams-kcsara, line 118: Maybe instead of maxHeight something like linesVisible would be more descriptive? MaxHeight immediately makes me think of pixels or em units, not lines.

src/components/OutputForm.tsx
```diff
@@ -47,17 +48,43 @@ export const OutputText = ({ label, value }: { label: string, value?: string })
   );
 }
 
-export const OutputTextArea = ({ label, value }: { label: string, value?: string }) => {
+export const OutputTextArea = ({ label, value, maxHeight }: { label: string, value?: string, maxHeight?: number }) => {
+
+  let content = (value !== undefined) &&  (
```
colincwilliams-kcsara, line 53: nit: can be const

src/components/OutputForm.tsx
```diff
@@ -47,17 +48,43 @@ export const OutputText = ({ label, value }: { label: string, value?: string })
   );
 }
 
-export const OutputTextArea = ({ label, value }: { label: string, value?: string }) => {
+export const OutputTextArea = ({ label, value, maxHeight }: { label: string, value?: string, maxHeight?: number }) => {
```
colincwilliams-kcsara, line 51: How does this change handle maxHeight being set, but the content is shorter than the maxHeight? Seems like the "Show More" button will still be visible, but do nothing?

valonm: yeah, the problem is we don't actually know how tall the content is. For example, if the description is a 200 word paragraph with no line breaks, the component only has 1 Typography element, but that element may wrap 20 lines. I think for now, the experience is fine -- I'd rather have the button show and not really be needed, versus, not implementing it and having someone enter a giant description that cannot be shrunk.

Maybe at some point we can figure out a way to infer the height of the content by the length or something.

src/app/(main)/EventPage.tsx
valonm: I'm going to use `rows` to align with the `textarea` prop name.

src/components/OutputForm.tsx
valonm: done

src/components/OutputForm.tsx
valonm: looking into this... it might be easier than I thought

colincwilliams-kcsara: Gotcha. And ya, looking it up it seems it can get messy to figure that out. Seems like there may be a way using react effects, which allow you to get the height of a rendered element, but agreed with getting this in now and fixing that later.

colincwilliams-kcsara (approved):

src/components/OutputForm.tsx
```diff
@@ -47,17 +48,43 @@ export const OutputText = ({ label, value }: { label: string, value?: string })
...
+    })
+  )
+
+  const [collapse, setCollapse] = useState<boolean>(!!rows);
+  const handleClick = () => {
+    if (!rows) { return; }
+    setCollapse(!collapse);
+  }
+
+  let linesToShow = collapse && (Math.max(rows ?? 0, 0) > 0) ? rows : undefined;
```
davidgut76, line 68: What's the point of the Math.max here? Would this be the same as 
collapse && ((rows ?? 0) > 0) ? ...

davidgut76 (approved):
What's the issue with showing partial rows? Doesn't that do a good job of indicating that there's more to see?

src/components/OutputForm.tsx
valonm: I don't want negative values.

src/components/OutputForm.tsx
valonm: I was able to address this in the commit I just made.

valonm:
> What's the issue with showing partial rows? Doesn't that do a good job of indicating that there's more to see?

I didn't like how it looked, and it was inconsistent depending on if/where line breaks were.

PR #37, opened 2023-07-04
Show Mission Status and Responder Counts

valonm:
- Derive Mission Status and Show on EventTile and Mission Page.
- Calculate responder count for various statuses.
  - Active: Count participants that are not in "Signed Out" status (includes "In Town")
  - Standy: Count participants in standby
  - Responding: Count participants in signedIn (these people are enroute TO the mission)
  - Checked-In: Count participants in Available and Assinged status (these people have arrived at the mission)

image
image

### DIFF
--- a/src/app/(main)/EventPage.tsx
+++ b/src/app/(main)/EventPage.tsx
@@ -7,9 +7,9 @@ import { useEffect, useState } from "react";
 import DeleteIcon from "@mui/icons-material/Delete";
 import { DataGrid, GridColDef, GridEventListener, GridRowsProp } from '@mui/x-data-grid';
 import { useAppDispatch, useAppSelector } from '@respond/lib/client/store';
-import { buildActivitySelector, isActive } from '@respond/lib/client/store/activities';
+import { buildActivitySelector, isActive, getActivityStatus } from '@respond/lib/client/store/activities';
 import { ActivityActions } from '@respond/lib/state';
-import { OrganizationStatus, Participant, ParticipatingOrg, ResponderStatus } from '@respond/types/activity';
+import { OrganizationStatus, Participant, ParticipatingOrg, ResponderStatus, isActive as isParticpantActive, isCheckedIn as isParticpantCheckedIn } from '@respond/types/activity';
 
 import { OutputForm, OutputLink, OutputText, OutputTextArea, OutputTime } from '@respond/components/OutputForm';
 import { StatusUpdater } from '@respond/components/StatusUpdater';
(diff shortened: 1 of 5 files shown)


src/app/(main)/EventPage.tsx
```diff
@@ -110,7 +126,11 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
             <OutputLink label="Map" value={activity.mapId} href={`https://sartopo.com/m/${activity.mapId}`} />
           </Box>
           <Box>
-            <OutputText label="Status" value={isActivityActive ? 'In Progress' : 'Complete'} />
+            <OutputText label="Mission Status" value={getActivityStatus(activity)} />
+            <OutputText label="Active Responders" value={Object.values(activity.participants).reduce(reduceActive, 0)}></OutputText>
+            <OutputText label="Standby" value={Object.values(activity.participants).reduce(reduceStandby, 0)}></OutputText>
+            <OutputText label="Responding" value={Object.values(activity.participants).reduce(reduceSignedIn, 0)}></OutputText>
```
colincwilliams-kcsara, line 132: Would this be better as "En Route"? Responding sounds very similar to Active Responders.

src/app/(main)/EventPage.tsx
```diff
@@ -110,7 +126,11 @@ export const EventPage = ({ eventId }: { eventId: string }) => {
             <OutputLink label="Map" value={activity.mapId} href={`https://sartopo.com/m/${activity.mapId}`} />
           </Box>
           <Box>
-            <OutputText label="Status" value={isActivityActive ? 'In Progress' : 'Complete'} />
+            <OutputText label="Mission Status" value={getActivityStatus(activity)} />
```
colincwilliams-kcsara, line 129: I'm concerned about overloading the table of mission details such that it becomes difficult to glance at. Perhaps it's worth putting the various numbers of responders in their own table or section for readability?

colincwilliams-kcsara (approved):
Leaving comments for follow-up, as I know you wanted this PR in for the deployment tonight. Nothing critical noted.

src/app/(main)/EventPage.tsx
valonm: I went with Responders because the individual status that you see when you're signed in is "Responding". I'm open to debate, just trying to align the terminology that we're using.

src/app/(main)/EventPage.tsx
valonm: Agreed at some point I think we'll want to put it in a sections and maybe make the entire sections collapsible.

PR #39, opened 2023-07-05
Add : Responders can provide ETA on Sign In

valonm:
### Added ETA Field and UI Componentry
- Add: participant.eta
- Add: ETA can be entered when signing in
- Add: ETA can be viewed on the Event Page Roster if the Responder is in status "Signed In"
- Responders can sign in with or without an ETA
- Not Implemented Yet - There is no way to edit ETA after you have signed in, except to sign out and back in again. I implemented this with the existing status change form. We'll need to implement a status edit form, which may be separate; or would require deeper refactor of the current form.

### Changed Roster Data Grid
- Re-implemented row height properly; it will be responsive to the density setting now.
  - I also noticed the previous fix wasn't working on Firefox, will need to check that with this. I don't have FF installed ATM.
- Added a custom toolbar to enable column show/hide and density selection.
  - ETA is hidden by default, because of screen real-estate on mobile. If we find that it is useful enough, maybe we re-evaluate.

image
image
image
image
image


src/app/(main)/EventPage.tsx
```diff
@@ -51,8 +52,20 @@ const Roster = ({participants, orgs, startTime}: {participants: Record<string, P
...
       return `${!isToday ? formatDate(o.value, 'yyyy-MM-dd ') : ''}${formatDate(o.value, 'HHmm')}`;
     }, flex: 1 },
+    { field: 'eta', headerName: 'ETA', flex: 1},
   ];
 
+  function CustomToolbar() {
+    return (
+      <GridToolbarContainer>
+        <GridToolbarColumnsButton />
+        {false && <GridToolbarFilterButton />}
```
davidgut76, line 62: What does this line do, and what's the "false &&" for?

src/app/(main)/EventPage.tsx
```diff
@@ -63,7 +76,20 @@ const Roster = ({participants, orgs, startTime}: {participants: Record<string, P
       hideFooter
       rowSelection={false}
       onRowClick={handleRowClick}
-      getRowHeight={() => 'auto'}
+      getRowHeight={({ densityFactor }: GridRowHeightParams) => {
+        return 52 * densityFactor;
```
davidgut76, line 80: What is magic about 52? Also I still don't understand what units these heights and widths are in. Is it pixels?

src/app/(main)/EventPage.tsx
```diff
@@ -51,8 +52,20 @@ const Roster = ({participants, orgs, startTime}: {participants: Record<string, P
       const isToday = new Date().setHours(0,0,0,0) === new Date(o.value).setHours(0,0,0,0);
       return `${!isToday ? formatDate(o.value, 'yyyy-MM-dd ') : ''}${formatDate(o.value, 'HHmm')}`;
     }, flex: 1 },
+    { field: 'eta', headerName: 'ETA', flex: 1},
```
davidgut76, line 55: This won't show a date, so it will be different from the "Time" column when the eta isn't today. Maybe that's ok, since this is only useful for a short time.

We could consider rendering it as a relative time, too, like "85m", or "-5m" if someone is overdue. Depends on which you think would be more useful.

davidgut76 (commented):
Mostly questions to help me understand the code.

src/app/(main)/EventPage.tsx
valonm: It's a shorthand for "if this is true, then render that". Normally there's a boolean evaluation in the first part, but I just put false to hide that element for now.

src/app/(main)/EventPage.tsx
valonm: Per the MUI documentation 52 is the default height of the data grid rows.

src/app/(main)/EventPage.tsx
valonm: Yeah, I think you might need to revisit this in the future, especially if there's ETAs that span over midnight, it will always assume the time is today. So if the mission spins up at 2300 and ETA is 0030 it will think the ETA is in the past. That's why I'm only rendering the time, for now. Maybe we add some logic to force the date to plus one If it's otherwise in the past, and then we could do relative time calculations. Which actually might be hopeful. Or maybe an option where it toggles.

valonm: Another thing that I was thinking, is that we might be able to graph the available responders as a forecast. So for example, we show current active responders and then we show active responders 30 minutes an hour and an hour and a half from now

src/app/(main)/EventPage.tsx
valonm: I think it's pixels. The default line height is 48.

PR #41, opened 2023-07-06
Automatically extract the map id from a full SARTopo link.

davidgut76:
This is a bit of a hack, and can easily be broken. Do we need better validation though? I could check for valid SARTopo/CalTopo links and only try to extract map ids from those, adding an error when there's something we don't understand. But I'm not sure how necessary it is, since ultimately we can't validate that a given map id is valid.


valonm:
I'm thinking something more along these lines:
~~~~
const scrubMapId = (value) => {
  if (!value) { return ''; }
  const regex = /m\/([a-z0-9]{4,})/gi;
  return Array.from(value.matchAll(regex), m => m[1])[0] ?? value;
}

console.log(scrubMapId(undefined));
// ""
console.log(scrubMapId('sartopo/m/45ACF0'));
// "45ACF0"
console.log(scrubMapId('45ACF0'));
// "45ACF0"
~~~~

This wouldn't be perfect, it only accounts for certain, expected inputs by the user -- basically a cal/sartopo link that includes a bit in the format "m/XXXXXX"

~~~~
console.log(scrubMapId('cat'));
"cat"
console.log(scrubMapId('/45ACF0'));
"/45ACF0"
console.log(scrubMapId('https://goo.gl/maps/cdw5kMFXCzqAzRzh7'));
"https://goo.gl/maps/cdw5kMFXCzqAzRzh7"
~~~~

davidgut76:
I'm not seeing much of an improvement in your code. I think the only extra thing it does is check that there are at least 4 numbers and letters after the slash. I'd vote for the simpler code. If you think it's important to try to validate the map id, I can add a check after the split.

valonm (approved):

PR #42, opened 2023-07-22
Custom status time

davidgut76:
Add a time control to allow users to set a specific time when changing their status. Also fix a related bug where the leg miles stayed in the form data and were added on every status change, even when the leg miles box was unavailable.

This works, but not quite how it should. Opening the status change form sets the current time in the input box. If a user doesn't make any changes, then the form data remains empty because the onChange handler is never called (setValue on line 194 being the important piece), and on line 64 in onSubmit we grab a new current time. Normally not a problem, but this would be weird if a user opened the dialog, waited an hour, then hit the button to change status. So ideally the form data would always match the time displayed in the UpdateStatus form. 

If I simply add a call to setValue in the StatusTimeInput function (say at line 201), I get a React warning after the submit, when it rerenders the StatusTimeInput: 
Warning: Cannot update a component (`Controller`) while rendering a different component (`StatusTimeInput`).

The warning only happens once per page refresh and overall it seems to work like I would want it to. But I'd like to understand what's going on here and possibly how to fix it. 
- why is it re-rendering StatusTimeInput after submit even before I've clicked the status button again?
- is there some event I could use to set the form data to be equal to the initial TextField value? 
- or is there some other right way to do this that I'm missing?

localhost_3000_mission_84c2e904-99ba-44f4-a2a6-200840ec4498(Pixel 5)


davidgut76:
Also I still need to remove the console.log statements...

mcosand:
Can you resolve the merge conflict?

davidgut76:
Ok, finally had time to get back to the this, figure out the merge and relearn about forms and state enough to get it working again. I still have the same questions, and I've updated the line numbers in the original comment.

davidgut76:
And now I've cleaned up the yarn.lock conflict, lint failures, and hopefully everything else. Now it's just the issue of my code.

davidgut76:
I think this is all working now, thanks to a bunch of Colin's help. One more thing I added since discussing with Colin is validation to make sure the status update times don't go back in time, since that would mess up the hours calculation and probably other stuff. 
localhost_3000_mission_07d72c1b-20f8-42ce-9a7c-5e48ef40b397(Pixel 5)

colincwilliams-kcsara (changes requested):

colincwilliams-kcsara (approved):

PR #45, opened 2023-09-23
Show start time on dashboard when the mission is scheduled in the future

valonm:
Fixing a minor pain point. When missions are in the future we should be able to see the start time on the dashboard, rather than clicking into the mission.

Added Start Time for future mission/activity, below status, in all 3 sections.

image

image

### DIFF
--- a/src/app/(main)/page.tsx
+++ b/src/app/(main)/page.tsx
@@ -6,13 +6,13 @@ import { Box, Button, Stack, Typography, Chip } from "@mui/material";
 //import styles from './page.module.css';
 import { useAppSelector } from '@respond/lib/client/store';
 import { canCreateEvents, canCreateMissions } from '@respond/lib/client/store/organization';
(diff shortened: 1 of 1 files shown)


davidgut76:
Totally agree that this is necessary - I was annoyed by it too. These tiles are already big, though, so here are some other thoughts about saving space:
- Is the "State #" useful on the main page? 
- Are the unit chips and Caltopo link also useful, or could they live only in details? 
- What about putting the start time after the title, like people have been doing manually? "Recovery Scene Training - Wed 2023-10-11 1900". That's pretty long, so it could also be just the date: "Recovery Scene Training - Wed 10-11"

mcosand (approved):

mcosand:
Merged this before we addressed David's comments. Good points, I think we can factor them into a UI refresh (adding tabs to details page, etc)

PR #47, opened 2023-09-26
Try different viewport setting to avoid zoomed-in view on navigation

mcosand:
The site will currently change viewport zoom while navigating the app. Try a more restrictive viewport setting to try to stop this from happening.


mcosand:
[Page zooms in on navigation](https://trello.com/c/kn0RrUS1/56-page-zooms-in-on-navigation)

colincwilliams-kcsara (approved):

PR #48, opened 2023-09-28
Setup effective(?) linting

mcosand:
Setup more stringent linting rules, including sorting `import` statements


colincwilliams-kcsara:
When does the linter run? And do builds automatically fail if the linter fails?

My strong preference is having it run automatically with, at minimum, production builds and failing the build. I'm personally a fan of having it run with dev builds as well, but I know that's more contentious as it slows down prototype-style development.

colincwilliams-kcsara (approved):

mcosand:
Linting runs as part of the `yarn lint` or `yarn build` tasks. It shouldn't run during `yarn dev`, but it's recommended that your IDE know about lint rules (Visual Studio Code with the ESLint extension should work).

There's a GitHub action that should block commits to `main` or merging PRs with failing builds.

PR #50, opened 2023-10-04
Create pages + APIs to list all missions/events

mcosand:
includes the `encoding` package, which is an optional but often installed dependency of some of our other packages.


mcosand:
[Overflow page for all activities](https://trello.com/c/yFS2VPLL/23-overflow-page-for-all-activities)

PR #51, opened 2023-10-25
Force line endings to LF, even on Windows

colincwilliams-kcsara:
Git on Windows defaults check out files with CRLF line endings and push them with LF line endings for consistency with other platforms. This generally works, but our linter is exclusively looking for LF endings. This means it's impossible to run the linter locally on Windows with the default Git behavior.

This changes the Git setting for this repository to, for all files, force Git to check out with LF line endings, guaranteeing compatibility with our linter for those on Windows. It requires no intervention from the user, meaning their global setting can stay untouched.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
-* text=auto eol=lf
+# Windows users: please set autocrlf to *false* for this repository for correct behavior. Do not set it here.
+#
+# From the command line inside this repo run: git config core.autocrlf false
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@
 - [Node.js] (https://nodejs.org/en) 16.x or higher
 - [Yarn](https://yarnpkg.com)
 
+**Windows users**:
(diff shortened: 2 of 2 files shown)


mcosand (changes requested):
What editor/IDE are you using? I've had `.gitattributes` cause problems in the past. Maybe I was using it wrong. My answer has been to use an IDE that's not dumb and can work in LF files.

I checked in a setting for VS Code (.vscode/settings.json) that should tell it to use LF only for new files. If you're using another editor, maybe we can find a similar setting.

colincwilliams-kcsara:
> What editor/IDE are you using? I've had `.gitattributes` cause problems in the past. Maybe I was using it wrong. My answer has been to use an IDE that's not dumb and can work in LF files.
> 
> I checked in a setting for VS Code (.vscode/settings.json) that should tell it to use LF only for new files. If you're using another editor, maybe we can find a similar setting.

Using VSCode as my editor, but for most of my Git operations I use the terminal. I suspect that's where I'm hitting issues. Every time I check out a new branch, Git converts everything to CRLF unless I change my global setting.

colincwilliams-kcsara:
Based on an offline discussion, updated README and .gitattributes to both direct the user to set autocrlf to false for this repository.

mcosand (approved):

PR #54, opened 2023-11-02
UI refactor / Desktop UI

mcosand:
UI refactor / desktop UI as discussed in slack


mcosand:
[Layout update](https://trello.com/c/gdR9wYbz/59-layout-update)

valonm:
The briefing panel grows to fill available space; this is a problem when there is a longer description.
image

I added a comfortable fixed width (400px) that is just shy of 50% at the mobile break point (900px.)
image
image

PR #56, opened 2023-11-12
Fix tags not loading until refresh

mcosand:
This bug actually turned out to be relatively hard to find. To use common actions/reducers between frontend and backend, there's a little bit of black magic in the composition of the client redux store and the server action handling.

On the client side, I had forgotten to include the reducer for the `participant/tag` action the server was sending, so it was silently being ignored.

Just before starting on this bug I was looking at adding some unit testing with Jest. I didn't come up with a good compile time check to make sure the client redux store included all of the necessary reducers, but I was able to come up with a Jest test that could do it (in `lib/client/store/__tests__/storeTests`). That test makes sure the reducer is included to be run, the other test smoke tests that the reducer logic is correct.

### DIFF
--- a/src/lib/client/store/activities.ts
+++ b/src/lib/client/store/activities.ts
@@ -1,9 +1,14 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { CaseReducer, createSlice } from '@reduxjs/toolkit';
+import { TypedActionCreator } from '@reduxjs/toolkit/dist/mapBuilders';
 
 import { ActivityActions, ActivityState, BasicReducers } from '@respond/lib/state';
 import { Activity, isActive as isParticipantStatusActive, ParticipantStatus, ParticipantUpdate } from '@respond/types/activity';
 
 import { RootState } from '.';
+
+export interface ReducerBuilderStub {
+  addCase<ActionCreator extends TypedActionCreator<string>>(actionCreator: ActionCreator, reducer: CaseReducer<ActivityState, ReturnType<ActionCreator>>): ReducerBuilderStub;
+}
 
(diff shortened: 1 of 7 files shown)


src/lib/client/store/activities.ts
```diff
@@ -1,10 +1,15 @@
...
+import { TypedActionCreator } from '@reduxjs/toolkit/dist/mapBuilders';
 
 import { ActivityActions, ActivityState, BasicReducers } from '@respond/lib/state';
 import { Activity, isActive as isParticipantStatusActive, ParticipantStatus, ParticipantUpdate } from '@respond/types/activity';
 
 import { RootState } from '.';
 
+export interface ReducerBuilderStub {
+  addCase<ActionCreator extends TypedActionCreator<string>>(actionCreator: ActionCreator, reducer: CaseReducer<ActivityState, ReturnType<ActionCreator>>): ReducerBuilderStub;
+}
```
colincwilliams-kcsara, line 11: Instead of creating a new type here, can't you just use the built-in type for the test builder: ActionReducerMapBuilder<ActivityState>

Or did you do this to be more explicit about which methods we're actually using from the main builder type?

colincwilliams-kcsara (approved):
Great find!

src/lib/client/store/activities.ts
mcosand: I wanted the interface so I could provide a stub in the test case. I didn't want to implement all the methods on `ActionReducerMapBuilder`, so I just pulled out the one method we're using. Good question, I hadn't actually verified that `ActionReducerMapBuilder` contained more than the one method.

PR #57, opened 2023-11-13
About md

valonm:
Pushing this alternate implementation to share this error that I can't figure out.

`/about/page`
image

Per the internet, this (or some variant of) should fix it. Some of them say to declare it as a generic React type... but `string` seems straightforward. Neither actually resolved the red underline. The page renders just fine; its just TS complaining.
```
// global.d.ts
declare module '*.md' {
  const value: string;
  export default value;
}
```
No Dice.


valonm:
I got this working and it made the text content easier to edit, but at the cost of control over other aspects of layout. I'm going to stick with my original implementation.

PR #58, opened 2023-11-13
No Explicit Any Warnings

colincwilliams-kcsara:
We had a number of `no-explicit-any` linter warnings that were bugging me. This PR fixes _almost_ all of them.

Most of them were fixed by finding the correct type instead of using Any. This included adding generic types or defining types.

A few of them required disabling that warning. I only did this when the code was clearly type fiddling, and did it on a per-line basis. I Whoever is writing code dynamically adding and removing properties should write the code to be resilient to whatever is passed to them.

The lone exception to the above is disabling the warning when the code had a comment explaining why it was doing the "wrong" thing. That comment made sense, so disabling the linter's warning continues that comment.

Finally, there was one warning I was unable to fix. It is clearly _not_ type fiddling, yet I could not get the types to be happy for the life of me. That was on line 60 of `lib\server\stateManager.ts`. I still wanted to get this PR out with the fixes I have. If you can fix that one, please do so.

### DIFF
--- a/src/components/LoginPanel.tsx
+++ b/src/components/LoginPanel.tsx
@@ -31,9 +31,7 @@ export default function LoginPanel() {
   }
 
   async function finishLogin(token: string) {
-    const res = (await Api.post<any>('/api/auth/google', {
-      token,
-    })) as AuthResponse;
+    const res = await Api.postBody<{ token: string }, AuthResponse>('/api/auth/google', { token });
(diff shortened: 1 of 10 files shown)


davidgut76:
Looks reasonable, though I can't say I understand all of the changes.

davidgut76 (approved):

PR #59, opened 2023-11-17
Roster Edit View

valonm:
WIP

I haven't added nav yet, so just go to a mission page and change "mission" to "roster" in the address bar.
image

There are some edge cases to work through, but basically it scrubs the timeline and removes inconsequential status transitions; only keeps the transitions that are relevant to building the roster.

image
image

Next steps:
- Test/Handle edge cases.
  - Turnarounds would result in Sign-in > Depart Base > Sign Out, currently. Planning to ignore Depart Base if there is no Arrive Base.
  - There might be edge cases related to the In Town status.
  - Need to test with the secondary status options.
- Make the values editable in the table.
- Have a way to save the roster entries to the activity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "@mui/icons-material": "^5.11.16",
     "@mui/material": "^5.11.12",
     "@mui/x-data-grid": "^6.0.3",
+    "@mui/x-date-pickers": "^6.18.2",
     "@react-oauth/google": "^0.8.0",
     "@reduxjs/toolkit": "^1.9.3",
     "@types/jest": "^29.5.8",
--- a/src/app/(main)/roster/[activityId]/page.tsx
+++ b/src/app/(main)/roster/[activityId]/page.tsx
@@ -1,0 +1,6 @@
+'use client';
+import { RosterEdit } from '@respond/components/activities/RosterEdit';
+
+export default function EditRoster({ params }: { params: { activityId: string } }) {
+  return <RosterEdit activityId={params.activityId} />;
+}
--- a/src/components/ClientOnly.tsx
+++ b/src/components/ClientOnly.tsx
@@ -1,5 +1,7 @@
 'use client';
 
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
 import React, { ReactNode, useEffect, useState } from 'react';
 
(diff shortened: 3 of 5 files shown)


valonm:
Revised Screenshot

image

colincwilliams-kcsara:
Looking more closely at this, I like where this is going. I know Bob has been loving having the timelines on people's roster cards, but this view will blow his mind! Also sets us up nicely for D4H integration.

Not going to review the code at this point as you still have work to do.

mcosand:
Can you walk through the use case(s) here? My understanding is that the screenshot from yesterday (includes "Roster Test" and Sign In / Arrive Base / etc columns) is that it's a view derived from the timeline. If this is an editable table, do the changes go back to the timeline? Are you filtering out changes in the timeline to show the derived table? If so, edits to the roster view can make the timeline invalid?

I'd assume that edits go to the source data, ie: we have an editor for the timeline, not the roster view.

valonm:
I think a call would be useful to talk though the questions/scenarios.

1. It creates a "Roster Entry" for the participant (represents a row on the paper roster)
2. Then, it starts at the oldest timeline event and looks for "Roster Stages" (Sign In, Arrive, Depart, Sign Out).
3. If the stage is not tracked, skip.
4. When it finds a Roster Stage, it checks if the Roster Entry already has a time in that stage
  * YES: skip
  * NO: Assign the timeline value to the Roster Entry for that stage. (there's some extra logic here to skip depart, if there was no arrive -- for turnaround use cases)
5. If the stage was "Sign Out" then it closes out that Roster Entry and creates a new one the next time a tracked statge is found.

> I'd assume that edits go to the source data, ie: we have an editor for the timeline, not the roster view.

It's not actually handling the save logic yet... but my thought was to make the timeline immutable and have the roster entries be editable. If a roster entry is added that becomes the new "last update" then we would push a timeline update; otherwise we would only parse net new timeline records the next time the roster edit view is loaded.

I'm not saving edits yet, but once I do, the plan is: Where there are _n_ Roster Entries already parsed, find the _n_ th Sign Out in the timeline and run the parsing logic from there forward starting with the most recent Roster Entry.

In this example, the timeline events with the tick marks are the ones that are pulled:
image

valonm:
Every time slot is now clickable and opens a dialog to edit the date/time. The actual save is not implemented yet -- pending further discussion on how/where to persist the data.

Updated Screenshots

image
image

PR #61, opened 2023-11-20
Roster Card: Incorrect Hours

colincwilliams-kcsara:
The calculated hours were incorrect on the roster cards due to counting some status change jumps multiple times. This change fixes it.

Added some comments for clarity while I was in there.

### DIFF
--- a/src/components/activities/RosterPanel.tsx
+++ b/src/components/activities/RosterPanel.tsx
@@ -147,7 +147,5 @@ function ParticipantHoursText({ participant }: { participant: Participant }) {
   }
 
   // Round to the nearest quarter hour.
-  const hours = Math.round(timeOnClock / 1000 / 60 / 15) / 4;
-
-  return <>{hours}</>;
+  return <>{Math.round(timeOnClock / 1000 / 60 / 15) / 4}</>;
 }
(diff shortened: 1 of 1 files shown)


src/components/activities/RosterPanel.tsx
```diff
@@ -136,14 +137,15 @@ function ParticipantHoursText({ participant }: { participant: Participant }) {
   }, [lastTimeline, latestTimeout, participant.firstname]);
 
   let timeOnClock = 0;
-  let timeout = latestTimeout;
+  let lastTime: number = latestTimeout;
   for (const t of participant.timeline) {
     if (isOnClock(t.status)) {
-      timeOnClock += timeout - t.time;
-    } else {
```
davidgut76, line 143: Basically other than the rename I think you just got rid of the "else"?

davidgut76 (approved):

src/components/activities/RosterPanel.tsx
colincwilliams-kcsara: Ya, I guess so! Conceptually went from always using the last time out in the subtraction to just using the last time. Practically that just meant removing the else.

PR #62, opened 2023-11-22
Standby Only Support

colincwilliams-kcsara:
Adds support for an activity to be marked as "Standby only". Similar to existing future missions this blocks responders from signing in, only allowing standby.

Already active responders are allowed to update their status as normal, allowing this toggle to be set for an active mission.

Additionally:

- Cleaned up some of the layout on the ActivityEditPage
- Fixed the existing "Run as Mock Mission" toggle to correctly display the existing value when editing an existing activity. It would always show as "off" when the edit page loaded, even if the value was stored as "on". This is because there is no "value" property on the component, and the "checked" property needed to be manually set to the value.

image
image
image


colincwilliams-kcsara:
[Standby Only](https://trello.com/c/vTa2MMkl/61-standby-only)

valonm (approved):

valonm:
Fixed a pre-existing issue with the alignment of these form elements:

Before
image

After
image

colincwilliams-kcsara:
Made a few final changes:

1. Changed the name of the field on Activity to "forceStandbyOnly" (from just "standbyOnly"). This reflects that an activity can be in a standby only mode through means other than the toggle; namely by being outside of the early sign-in window. The setting on the Activity is forcing the activity into that state regardless of other logic.
2. Tweaked the grid (screenshots below) so that the cancel/save buttons float up to the top on desktop.
3. Fixed a bug I just discovered where, when standbyOnly is forced, responders on Standby would be presented with the non-standby only options. This means they were presented with the sole option of "Sign In", exactly the opposite of what's desired.

image
image

valonm:
Seems like something isn't right about how the switch state is being updated.

image

Also found an edge case which is probably related to how we are determining the valid next statuses. Example, if I am already in "Demobilized" status (Departed Base, or Turned Around), and the mission goes to Standby Only, I have options for Sign Out, Sign In or Arrive Base. This is not a normal situation we would end up in probably, but technically the options should be Sign Out or Standby under these conditions. There are probably other edge cases in that "next status" logic that needs to use the "future mission" logic when forceStandby is true.

valonm:
I'm good with the current state of this update. We can address the edge cases for "next status options" logic separately.

colincwilliams-kcsara:
Ok, I looked into it and understand the consequences of that error. I do not believe it's worth fixing.

The source of the error is the bound value in defaultValues for the form being undefined. I was able to reproduce it with the "Run as Mock Mission" switch if I deleted the "asMission" field in the database (manually).

Given it only impacts past activities, and this new setting is highly, highly unlikely to be changed on a past activity, adding code to handle it seems unnecessary. Additionally, it appears there is no functional impact to the app should it be encountered.

That's a super interesting edge case. Agreed on it not blocking this PR, but will start working on a fix ASAP to clean it up.

PR #64, opened 2023-11-23
Soft Delete

colincwilliams-kcsara:
Stamps "deleted" activities with a removal time rather than actually deleting them from the database. This is transparent to the client - as far as the client is concerned, the activity no longer exists.

Additionally, since soft-deleted activities aren't retained in memory, they have no impact on memory usage or list operations (e.g. always having to filter them out). The only performance impact is at server start, where they are filtered out when priming the server's in-memory state.

### DIFF
--- a/src/lib/server/stateManager.ts
+++ b/src/lib/server/stateManager.ts
@@ -11,6 +11,8 @@ import { ActivityActions, ParticipantUpdateAction } from '../state/activityActio
 
 import { getServices } from './services';
 
+type DatabaseActivity = Activity & { removeTime?: number };
+
 export interface ActionListener {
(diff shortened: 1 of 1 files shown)


colincwilliams-kcsara:
[Soft Delete](https://trello.com/c/NVlCSZ3j/62-soft-delete)

valonm (approved):

mcosand (approved):
Looks great.
When we get to partially connected clients, there will be complexity in resolving if a client is talking about a new mission, or a mission that's been marked as removed. That sounds like a later problem, though.

colincwilliams-kcsara:
Yes, was thinking of that as well. Agreed on it being a later problem!

PR #65, opened 2023-11-26
Mobile UI Adjustments

valonm:
A few additional adjustments to the mobile view.
- ~Default the mobile view to the roster panel.~
- Position `StatusUpdater` in the footer component so that it is available in all mobile views.
- Add optional `fullWidth` property to `StatusUpdater`.

image
image
image
image

### DIFF
--- a/src/components/SplitButton.tsx
+++ b/src/components/SplitButton.tsx
@@ -12,7 +12,7 @@ export function splitOptionsFromStrings(options: string[]) {
   return options.map((o) => ({ id: o, text: o }));
 }
 
-export function SplitButton<K extends string | number | EnumMember, T extends IdOption<K>>({ options, selected, onClick }: { options: T[]; selected?: K; onClick?: (value: K) => void }) {
+export function SplitButton<K extends string | number | EnumMember, T extends IdOption<K>>({ fullWidth, options, selected, onClick }: { fullWidth?: boolean; options: T[]; selected?: K; onClick?: (value: K) => void }) {
   const [open, setOpen] = useState(false);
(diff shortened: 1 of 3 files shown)


colincwilliams-kcsara:
Love the status button position!

My only concern with this PR is switching the default tab to the roster without addressing the activity description. With the description remaining solely in the Briefing tab it's now two taps away, and I think it's _highly_ unlikely a responder will go find it.

We need to solve that problem - IMO by making the description accessible directly from the main activity list, but I'm very open to other ideas - before we change the default tab.

At the end of the day folks who want the roster tab will get to it. I don't believe folks who are less invested in/interested in the app (casual responders) will seek out the description if it's buried.

valonm:
Hmm... maybe I am using it different than others, but the roster is always the first thing I want to see. I'd inverse your comment; At the end of the day folks who want the description will get to it.

valonm:
I've removed the default tab change from this PR.

colincwilliams-kcsara:
Oh ya, I use it the exact same way as you and I really want the primary tab to be the roster tab.

But my belief, and maybe I'm wrong, is that people looking at the roster view are leaders in their unit. If not by title, then by practice. They actively seek out information and are more plugged in.

Folks who don't care about the roster are, in my mind, less likely to be plugged in and actively seeking out information. Thus they won't seek out the description and learn what it has to say - which may be important/useful.

So that's where I'm coming from on it. Sounds good, though. Will approve without the default tab change.

colincwilliams-kcsara (approved):

PR #66, opened 2023-11-26
Implement client preferences using local storage.

valonm:
Implements Local Storage to save client preferences.
- Initial use case is `defaultMobileView`.
  - If there is no saved preference, it defaults to "Briefing"
  - If there is a saved preference, the initial state is set accordingly

image
image

### DIFF
--- a/src/components/AppMenu.tsx
+++ b/src/components/AppMenu.tsx
@@ -8,47 +8,67 @@ import * as React from 'react';
 import { useAppDispatch, useAppSelector } from '@respond/lib/client/store';
 import { AuthActions } from '@respond/lib/client/store/auth';
 
+import { PreferencesDialog } from './Preferences';
+
 export function AppMenu() {
   const dispatch = useAppDispatch();
   const { userInfo } = useAppSelector((state) => state.auth);
   const [menuAnchor, setMenuAnchor] = React.useState<HTMLElement | null>(null);
   const handleMenu = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => setMenuAnchor(event.currentTarget);
   const handleClose = () => setMenuAnchor(null);
+
+  const [openPreferences, setOpenPreferences] = React.useState(false);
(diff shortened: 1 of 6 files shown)


mcosand:
I want to move more towards putting weather/driving directions/etc in this app instead of across the pager. I think for most people, the briefing screen is the correct default view. I like this approach a lot better so that mission managers/response coordinators can save a tap or two for their work.

I'm not excited about components reaching up and around to read preferences out of the preferences UI component. Trying to think of an easy yet consistent way to manage preferences. I think it ends up in the redux store, but I don't have strong opinions about how that's done, or if the preferences store is some non-redux object in the lib/client directory. Maybe a small but forgivable sin to reach into the preferences component for now?

colincwilliams-kcsara:
Also love the new approach! A setting is great.

Agreed with Matt that the Redux store is the correct place for the preferences to live. By not setting the "sync" property on the actions, the preferences will stay local (not synced to the server and, ultimately, other clients) but will follow the established patterns the rest of the data in the app uses.

valonm:
Thinking I should have left the `localStorage` update in the save method on the Preferences Dialog.

colincwilliams-kcsara:
By my understanding of redux, side effects should go in sagas, thunks, or an equivalent. Essentially handled by a middleware that kicks off side effects in response to a redux store action happening.

You can see an example of this lib/client/sync.ts, where the app persists the activity state to disk. That would be an exact match to what you're doing.

(We aren't using sagas or thunks in our app right now. We're using a simpler variant. But the general idea is the same.)

valonm:
I think this is a comparable use case in `LoginPanel`, and it just does it right before the dispatch.

image

colincwilliams-kcsara:
Ah that's fair. I'm good with mimicking LoginPanel. I still believe the middleware solution is the formally "correct" option, but for a simple case like this - especially a synchronous case - I'm good with it.

colincwilliams-kcsara (approved):

PR #67, opened 2023-11-29
Relative times

davidgut76:
Use relative date/times by default when the time is within 1 calendar day of today. Standardize our time output by moving more places to use OutputTime rather than doing custom date formatting. Delete Roster.tsx since I believe it is no longer used (thus eliminating its usage of date-fns).

A couple of places I couldn't/didn't fix to use OutputTime:
- in the UpdateStatusForm error message, because OutputTime returns a component and this needs just a string. I could refactor code to standardize this too, but then we could get things like "Cannot be earlier than previous status change at yesterday at 1345", which reads a bit odd. So it seems ok to just switch it to use our "standard" absolute time format. 
- also in UpdateStatusForm where we parse/format into the TextBox format, which is very specific and annoying.
- FormUtils.ts, which is only used by ActivityEditPage to split/combine date times. My preference would be just to use a date picker and not reinvent the wheel here, but that's a different discussion.

We could go further with relative dates if we wanted (say within 1 week).

localhost_3000_mission_3094bd61-f825-4305-9ca0-1184cdb968d2(Pixel 7)
localhost_3000_mission_3094bd61-f825-4305-9ca0-1184cdb968d2(Pixel 7) (1)
localhost_3000_mission_3094bd61-f825-4305-9ca0-1184cdb968d2(Pixel 7) (2)


valonm:
I like the one day window for relative time on event cards.

I don't really like it on timeline; there is other work/discussion happening around timeline, too. Maybe don't make changes there right now for that reason alone.

My 2 cents, I prefer the plain date and time on the timeline view because that's what is needed when entering rosters in the database and I don't want mental gymnastics between entries.

colincwilliams-kcsara:
Agreed with Valon about the timeline only having absolute dates.

For the case where you just need a string, not the full component, can't you just export a method formatDate(allowRelative: boolean)? That would return the string. The component would use it, but if the component isn't needed the method could be used directly.

davidgut76:
I restored the roster view to always show absolute time, but kept the OutputTime component in there, as it's cleaner. OutputTime now only auto-selects the view if `relative` is explicitly undefined. 

Regarding exporting a method to get the string, yes I could do that (and already tried it), but it's a bunch of extra goop just to avoid `formatDate(lastStatusChangeTime, 'EEE yyyy-MM-dd HHmm')`, and getting relative dates in the error message isn't a great experience. So my point is that it's not really worth the code complexity given the rarity of the error message and that we'll still need `formatDate` in that file for the `TextBox`. Instead I moved the string string defining our "standard" absolute format and the text box format into RelativeTimeText, so at least those are centralized.

colincwilliams-kcsara (approved):

src/components/OutputForm.tsx
```diff
@@ -144,5 +145,12 @@ export const OutputTime = ({ label, time, relative }: { label: string; time?: nu
     };
   });
 
-  return <OutputField label={label}>{time && <RelativeTimeText time={time} baseTime={nowTime} defaultToTime={!relative} />}</OutputField>;
+  let useRelative = relative ?? false;
```
valonm, line 148: This seems unnecessary, `undefined` is already `falsey`

src/components/OutputForm.tsx
```diff
@@ -144,5 +145,12 @@ export const OutputTime = ({ label, time, relative }: { label: string; time?: nu
...
+  if (relative == undefined && time) {
```
valonm, line 149: The `relative` property should be an opt-in, not an opt-out.

I don't like this approach. Omitting the property should not mean it applies. For example, if we had a `bold` prop then `<OutputTime bold />` would make it bold and `<OutputTime />` would not be bold; we shouldn't have to do `<OutputTime bold=false />`.

A boolean prop is either present or it is not; present should be "do the thing" and not present should be "don't do the thing".

src/components/OutputForm.tsx
```diff
@@ -144,5 +145,12 @@ export const OutputTime = ({ label, time, relative }: { label: string; time?: nu
...
+    // If the time is within 1 day of today, use relative time.
+    const dateDiff = differenceInCalendarDays(new Date(), new Date(time));
+    if (Math.abs(dateDiff) <= 1) useRelative = true;
```
valonm, line 152: This doesn't feel like the right place to do this. Maybe `RelativeTimeText` should have a prop like relativeDaysRange that applies this logic so that individual implementations can just pass an integer.

`<RelativeTimeText time={time} baseTime={nowTime} relative={useRelative} relativeDaysRange=1 />`

src/components/activities/RosterPanel.tsx
```diff
@@ -86,7 +86,9 @@ export function ParticipantDialog({ open, participant, activity, onClose }: { op
                   <TableRow key={t.time}>
                     <TableCell>{activity.organizations[t.organizationId].rosterName ?? activity.organizations[t.organizationId].title}</TableCell>
                     <TableCell>{getStatusText(t.status)}</TableCell>
-                    <TableCell>{formatDate(t.time, 'EEE yyyy-MM-dd HHmm')}</TableCell>
+                    <TableCell>
+                      <OutputTime label="" time={t.time} relative={false}></OutputTime>
```
valonm, line 90: Not a fan of using the Output Form Fields like this. They apply unnecessary styling, here.

This could just be a RelativeTimeText:
```
<TableCell>
  <RelativeTimeText time={time} baseTime={new Date().getTime()} />
</TableCell>
```

valonm (changes requested):

src/components/OutputForm.tsx
davidgut76: Switched to an enum

src/components/activities/RosterPanel.tsx
davidgut76: Was there different styling that happened? I didn't notice it if so. After merging your changes, there's actually no need to use RelativeTimeText here at all anyway. I see that you're using a different DateTimePicker for the roster edit versus what we have in the update status dialog. We should probably be consistent with which DateTimePicker we use.

src/components/RelativeTimeText.tsx
```diff
@@ -24,24 +25,45 @@ const locale: Locale = {
...
       text = text[0].toLocaleUpperCase() + text.substring(1);
     }
   } else {
-    text = formatDate(time, 'EEE yyyy-MM-dd HHmm');
+    text = formatDate(time, AbsoluteDateFormat);
   }
+  return text;
+};
+
+export const RelativeTimeText = ({ time, baseTime, relative, lowercase }: RelativeTimeTextProps) => {
```
valonm, line 57: Give baseTime a default so we don't have to pass in `new Date().getTime()` everywhere:
```
... { time, baseTime = new Date().getTime(), relative, lowercase } ...
```

valonm (changes requested):

src/components/RelativeTimeText.tsx
```diff
@@ -24,24 +25,45 @@ const locale: Locale = {
...
+  let useRelativeDefault = relative == RelativeStyle.Relative;
+  if (relative == RelativeStyle.Auto && time) {
+    // If the time is within 1 day of today, use relative time.
+    const dateDiff = differenceInCalendarDays(new Date(), new Date(time));
+    if (Math.abs(dateDiff) <= 1) useRelativeDefault = true;
+  }
+
+  const [useRelative, setUseRelative] = React.useState<boolean>(useRelativeDefault);
```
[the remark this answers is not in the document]
valonm: `useState`, `useEffect`, etc. they all start with `use`. If `relative` causes a conflict then maybe call it `isRelative` -- which would be more common convention for Boolean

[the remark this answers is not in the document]
valonm: Also... I thought you had changed the prop name. If that wasn't in the scope of your changes, then I'm fine leaving as-is.

src/components/OutputForm.tsx
```diff
@@ -138,7 +138,7 @@ export const OutputLink = ({ label, value, href, target }: { label: string; valu
   );
 };
 
-export const OutputTime = ({ label, time, relative }: { label: string; time?: number; relative?: boolean }) => {
+export const OutputTime = ({ label, time, relative = RelativeStyle.Auto }: { label: string; time?: number; relative?: RelativeStyle }) => {
```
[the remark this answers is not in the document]
valonm: I'm fine with 1 day. We can make it dynamic later if we have a use case.

src/components/activities/ActivityListPage.tsx
```diff
@@ -25,7 +25,9 @@ function ActivityList({ activities }: { activities: Activity[] }) {
             <td style={{ paddingRight: '1em' }}>
               <Link href={getActivityPath(a)}>{a.title}</Link>
             </td>
-            <td>{formatDate(a.startTime, 'yyyy-MM-dd EEE HHmm')}</td>
+            <td>
+              <OutputTime label="" time={a.startTime}></OutputTime>
```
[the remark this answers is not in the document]
valonm: Absolute seems like the right answer here since the primary use case is to look for events that have expired from the main page, i.e. more than 1 day ago in any case

src/components/RelativeTimeText.tsx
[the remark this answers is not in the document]
davidgut76: Changed all of the useRelatives to isRelative.

PR #69, opened 2023-11-30
fix bottom nav filler on mobile

valonm:
The bottom nav requires more reserved height. The nav tabs are 56 pixels and the status updater is 68.5. updated the filler to 124.5 pixels.

### DIFF
--- a/src/components/activities/MobileActivityPage.tsx
+++ b/src/components/activities/MobileActivityPage.tsx
@@ -16,6 +16,8 @@ import { ManagerPanel } from './ManagerPanel';
 import { ParticipatingOrgChips } from './ParticipatingOrgChips';
 import { ParticipantDialog, RosterPanel, RosterRowCard } from './RosterPanel';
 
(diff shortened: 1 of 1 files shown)


valonm:
image

src/components/activities/MobileActivityPage.tsx
```diff
@@ -16,6 +16,8 @@ import { ManagerPanel } from './ManagerPanel';
 import { ParticipatingOrgChips } from './ParticipatingOrgChips';
 import { ParticipantDialog, RosterPanel, RosterRowCard } from './RosterPanel';
 
+const MOBILE_FOOTER_HEIGHT_PIXELS = 124.5;
```
colincwilliams-kcsara, line 19: Maybe do this as a calculation summing the two heights so it's clear where the number came from? Or at least note the two heights in a comment.

colincwilliams-kcsara (approved):
Thanks for the quick turnaround!

PR #73, opened 2023-12-05
Require sign out on org change.

valonm:
Changes the actions logic in Status Updater to require signing out of the previous unit before signing into another unit. This ensure continuity in the timeline when users change orgs.

Previously, situations like this were possible:
- Sign In (Unit 1)
- Sign Out (Unit 1) <-- due to automated sign out.
- Arrive Base (Unit 2) <-- new roster status in timeline with no corresponding sign in.
- Depart Base (Unit 2)
- Sign Out (Unit 2)

~We could potentially automate the sign in; but there are a lot of edge cases if the user can move to any potential status and we just sign them out automatically, sign them in automatically, then move them to that arbitrary status. I think it is better if we just have them sign out and then choose a new sign in status. I think in the unusual case a person is changing units mid-mission a little extra friction is okay to make sure we capture good data.~

### [UPDATED]

I have updated the logic for both the status updater and the Participant Update Reducer to better-handle "changing unit" transitions. There are 4 use cases:
- Switching while in a "Responding" status => Sign Out of the previous org, Sign In to the new org.
- Switching while in "Standby" status => Sign Out of the previous org, Standby to the new org.
- Switching while in "Remote" status => Sign Out of the previous org, sign in as In Town to the new org.
- Signing Out while in an "Active" status => Only Sign Out of the previous org. (this was an unhandled edge case prior to this change)

Sign Out of previous unit is available in all cases where the user signed into a different unit portal.

## Switching from a "Responding To Base" status
image
image

## Switching from "Standby"
image
image

## Switching from "In Town"
image
image

## Signing out from Previous Unit
image
image

## Final Roster
image

### DIFF
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -2,7 +2,7 @@ import { useState } from 'react';
 
 import { useAppSelector } from '@respond/lib/client/store';
 import { defaultEarlySigninWindow, isFuture } from '@respond/lib/client/store/activities';
-import { Activity, isActive, isResponding, ParticipantStatus } from '@respond/types/activity';
+import { Activity, getOrganizationName, isActive, isResponding, ParticipantStatus } from '@respond/types/activity';
 import { MyOrganization } from '@respond/types/organization';
 import { UserInfo } from '@respond/types/userInfo';
 
@@ -119,7 +119,7 @@ const StatusUpdaterProtected = ({ activity, current, fullWidth, user, thisOrg }:
   const currentOrgId = participant?.organizationId;
   let actions;
   if (isActive(current) && currentOrgId !== lastOrgId) {
-    const lastOrgName = activity.organizations[lastOrgId].rosterName ?? activity.organizations[lastOrgId].title;
+    const lastOrgName = getOrganizationName(activity, lastOrgId);
     actions = [{ id: 7, newStatus: ParticipantStatus.SignedOut, text: `Sign Out from ${lastOrgName}` }];
   } else {
     actions = getStatusOptions(current, activity.startTime, activity.forceStandbyOnly, activity.earlySignInWindow);
--- a/src/components/StatusUpdater/index.tsx
+++ b/src/components/StatusUpdater/index.tsx
@@ -116,11 +116,24 @@ const StatusUpdaterProtected = ({ activity, current, fullWidth, user, thisOrg }:
   }
 
   const lastOrgId = participant?.timeline[0].organizationId;
-  const currentOrgId = participant?.organizationId;
+  const currentOrgId = thisOrg.id;
   let actions;
   if (isActive(current) && currentOrgId !== lastOrgId) {
-    const lastOrgName = getOrganizationName(activity, lastOrgId);
-    actions = [{ ...statusTransitions.signIn, text: `Switch from ${lastOrgName}` }];
+    let transition;
+    switch (current) {
+      case ParticipantStatus.Remote:
+        transition = statusTransitions.inTown;
+        break;
+      case ParticipantStatus.Standby:
+        transition = statusTransitions.standBy;
+        break;
+      default:
+        transition = statusTransitions.signIn;
+    }
+    actions = [
+      { ...transition, text: `Switch from ${getOrganizationName(activity, lastOrgId)}` },
+      { ...statusTransitions.signOut, text: `Sign Out from ${getOrganizationName(activity, lastOrgId)}` },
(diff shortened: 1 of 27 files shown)


src/components/StatusUpdater/index.tsx
```diff
@@ -114,7 +115,15 @@ const StatusUpdaterProtected = ({ activity, current, fullWidth, user, thisOrg }:
     setConfirming(true);
   }
 
-  const actions = getStatusOptions(current, activity.startTime, activity.forceStandbyOnly, activity.earlySignInWindow);
+  const lastOrgId = participant?.timeline[0].organizationId;
+  const currentOrgId = participant?.organizationId;
+  let actions;
+  if (currentOrgId !== lastOrgId) {
```
valonm, line 121: This needs a condition on current status not equal to signed out or not responding.

src/components/StatusUpdater/index.tsx
```diff
@@ -114,7 +115,15 @@ const StatusUpdaterProtected = ({ activity, current, fullWidth, user, thisOrg }:
     setConfirming(true);
   }
 
-  const actions = getStatusOptions(current, activity.startTime, activity.forceStandbyOnly, activity.earlySignInWindow);
+  const lastOrgId = participant?.timeline[0].organizationId;
+  const currentOrgId = participant?.organizationId;
+  let actions;
+  if (isActive(current) && currentOrgId !== lastOrgId) {
+    const lastOrgName = activity.organizations[lastOrgId].rosterName ?? activity.organizations[lastOrgId].title;
+    actions = [{ id: 7, newStatus: ParticipantStatus.SignedOut, text: `Sign Out from ${lastOrgName}` }];
```
colincwilliams-kcsara, line 123: Magic number 7 for the ID. Should that be an enum value?

colincwilliams-kcsara, line 123: When the sign out goes through, is it stamped as the correct org (their previous one)? I'm not seeing anything in the action specifying which org the action applies to, so presumably it would apply to their current org?

valonm: hmm... that's a good call, let me look into that.

valonm: fixed

valonm: fixed

src/components/StatusUpdater/index.tsx
```diff
@@ -114,7 +115,28 @@ const StatusUpdaterProtected = ({ activity, current, fullWidth, user, thisOrg }:
...
+    let transition;
+    switch (current) {
+      case ParticipantStatus.Remote:
+        transition = statusTransitions.inTown;
+        break;
+      case ParticipantStatus.Standby:
+        transition = statusTransitions.standBy;
+        break;
+      default:
+        transition = statusTransitions.signIn;
```
colincwilliams-kcsara, line 131: Could easily be missing something, but in your PR description you say that in an active state the only thing that happens is the user is signed out of the old org. But this status transition indicates they're being moved to signed in, no?

valonm: Yeah, the second half of the description explains some use cases where I decided to handle the sign in for the user, after all.

colincwilliams-kcsara: Right, that's what I'm referring to. In that paragraph you say:

> Signing Out while in an "Active" status => Only Sign Out of the previous org. (this was an unhandled edge case prior to this change)

But the code is reading to me that the user will be placed into a signed in state, not signed out.

valonm: Ah, yeah, at this point the user hasn't actually changed their status yet. This is just just determining which transition options are _available_. Specifically, which sign-in option should be available if you are currently active with Unit A, but logged into the portal of Unit B. The sign out option is always available and is appended on Line 135, just below.

The logic that actually handles the sign-in/out is in the activity reducers. Basically, if the new status is from a different org, it signs them out of the previous org -- then if the new status is an active status it continues with the regular update in the current org. Sign out aborts the change in the current org because we replace it with a sign out for the previous org.

From the examples above, its the last scenario -- I am signed into CRT, but I'm logged into ESAR's Portal.
- I have a "Switch from CRT" option, which would sign me out of CRT and back in (to "In Town" in this example) as ESAR.
- I also have a "Sign Out from CRT" option that will just sign me out from the other org.

image

In this scenario I chose the secondary option from within the ESAR portal, and I am now signed out of CRT, even though I took the action from the ESAR portal.
image

PR #74, opened 2023-12-07
Timeline UI Overhaul

valonm:
Refactor the timeline in the participant dialog to a tile-based solution that works much better on mobile. Also, some minor enhancements to the timeline edit functionality to include a cancel button.

## Desktop
image
image

## Large Mobile Phone
image
image

## Small Mobile Phone
image
image

### DIFF
--- a/src/types/activity.ts
+++ b/src/types/activity.ts
@@ -69,6 +69,10 @@ export function getStatusText(status: ParticipantStatus) {
   return STATUS_TEXT[status];
 }
 
+export function getOrganizationName(activity: Activity, organizationId: string) {
+  return activity.organizations[organizationId].rosterName ?? activity.organizations[organizationId].title;
+}
+
 export function isActive(status: ParticipantStatus) {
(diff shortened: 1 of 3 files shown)


src/types/activity.ts
```diff
@@ -69,6 +69,10 @@ export function getStatusText(status: ParticipantStatus) {
   return STATUS_TEXT[status];
 }
 
+export function getOrganizationName(activity: Activity, organizationId: string) {
+  return activity.organizations[organizationId].rosterName ?? activity.organizations[organizationId].title;
```
colincwilliams-kcsara, line 73: I noticed you have this same logic, but inline rather than as a selector, in your other PR. Can you make sure this selector is used there as well? However you want to do that.

colincwilliams-kcsara (approved):

src/types/activity.ts
valonm: Yeah, I can update the other PR after this is merged. The dependency doesn't exist yet, obviously.

PR #75, opened 2023-12-08
Find Members (Need Help)

valonm:
I'm trying to figure out how to call D4H to query members by name. This was super easy to setup and test in Postman, so I know the D4H endpoint is good. It actually seems like it is failing to even call the internal API that I set up. What am I missing here?

image

image

### DIFF
--- a/src/app/(main)/test/page.tsx
+++ b/src/app/(main)/test/page.tsx
@@ -1,16 +1,19 @@
 'use client';
 
-import { Button, Stack, TextField } from '@mui/material';
+import { Button, Card, CardContent, Stack, TextField, Typography } from '@mui/material';
 import { useState } from 'react';
 
 import { ToolbarPage } from '@respond/components/ToolbarPage';
+import { apiFetch } from '@respond/lib/api';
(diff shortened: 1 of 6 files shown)


mcosand:
What card are you working on? I have https://trello.com/c/SvDRUIVT/13-adding-someone-else-to-the-roster and have member search working on a branch

mcosand:
pushed a couple changes to get this working

valonm:
Ok, so I got this figured out and turned it into a member directory type functionality (I'd still do a bit more work to polish it up and put it on a proper page). Not sure what type of reception this would get -- people are finicky about this type of stuff -- but this would be a really, really handy feature to have, and save me from having to log into D4H all the time.

Thoughts?

image

image

image

mcosand:
Glad you got things working.

A general membership directory feels like scope creep to our mission response app. Our current member management tool is a pain to use, and you have the option of creating an easier interface here, so you're going to be more inclined to consider it here. I wouldn't doubt that after adding a search tool, someone would want to add the ability to see someone's mission participation from that view, etc.

I'm not opposed to a façade in front of D4H, but would highly suggest that it get split out as a separate project. Especially if we're moving to containers.

PR #76, opened 2023-12-13
Locations

valonm:
This is an enhancement to add more-robust location data to activities which will enable planned functionality such as generating navigation links and calculating ETAs.

**This initial implementation**
- Enables locations to be selected/created on the Activity Edit page.
- Location data is synced to the Redux store and persisted into a new MongoDB collection.
- Single-use locations can be created for places that we wont be returning to; evidence searches, for example.

**Planned for future enhancement:**
- A "Locations" page where locations can be created/updated/deleted.
- Ability to edit full location details; initially I envision having some additional data like a description and offline driving directions.

## Deployment
A new mongodb collection is required: `locations`
image

Activities already store location as an object - no change required.
image

## New Component: LocationAutocomplete
Implements the Autocomplete component to render a searchable list of locations.
image

## New Component: NewLocationDialog
This dialog is presented when creating a new location from the Mission Edit page. This will give us enough information to generate a "Navigate" link. It's not intended to be the full location details; just a quick way to create the important bits.
* Enables quick creation of a location with as little as a name.
* If "Save for Later" is unchecked then the location is not persisted to the database. (Single Use Locations)
image

## Large Mobile
image

## Small Mobile
image


valonm:
Ok, so I've been working on implementing a Locations Manager page and I've uncovered an issue with how the websocket stuff works, which is results in a bunch of errors on the server side. Every action gets passed to both handlers -- location actions throw in the activity side and visa versa.

I'm not sure what the right pattern is to deduplicate the ActivityAction and LocationAction stuff, but it seems like that is probably the right answer, here.

image

valonm:
I'm also thinking that temporary locations should not be pushed to the store; they already don't go to the database.

1. Once they are in the store, they persist there until the server is rebooted. I don't think we really want that.
2. I had originally pushed them there because I thought it would be necessary for editing them -- but now I'm thinking that maybe the right solution is just to edit the location as it is stored within the activity. TBD on specific workflow.

valonm:
I've done some refactoring to implement `useForm` and make the Location edit/create form re-usable.

image

I've also added a locations admin page where locations can be edited/deleted (eventually, I'm having issues with MongoDB updates...)

image

As noted above, I am having issues implementing database updates. I can create locations just fine, but I am getting an error on reload/update:
~~~
unhandledRejection: MongoServerError: After applying the update, the (immutable) field '_id' was found to have been altered to _id:
~~~

I haven't been able to figure out what is causing this. Nothing is modifying the `_id` field. Everything appears to be implemented the same as the activity updates, which are not having this problem.

mcosand:
I've started a branch at `mcosand/locations` to merge the activity and location action handling. We should coordinate progress on the branches. I'm hoping to get more time on this this weekend.

src/components/locations/LocationEditDialog.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useAppDispatch } from '@respond/lib/client/store';
+import { LocationActions } from '@respond/lib/client/store/locations';
+import { createNewTemporaryLocation, Location } from '@respond/types/location';
+
+import { LocationEditForm } from './LocationEditForm';
+
+export function LocationEditDialog({ location = createNewTemporaryLocation(), open, onSubmit, onClose }: { location?: Location; open: boolean; onSubmit: (location: Location) => void; onClose: () => void }) {
+  const dispatch = useAppDispatch();
+  const handleSubmit = (location: Location) => {
+    if (location.active) {
```
valonm, line 12: Something more-robust will be needed, here, to handle use cases for editing a location for this activity only.

valonm:
I've pushed some updates to finalize the UI Workflows for Locations:
* Saved Location Manager: `/admin/locations`
  * Saved Locations can be created/updated/deleted
* Activity Location Edit:
  * The Activity Edit Page now has both edit and create location options, via the `LocationEditDialog`. From here, the `activity.location` can be edited. By default, location edits in the activity page are only for that activity.
  * If `activity.location` is a Saved Location, then there is an option to also update the Saved Location.
  * If `activity.location` is not a Saved Location, then there is an option to create a Saved Location.

When a location is selected, an edit option is available:
image

Update Option in Dialog:
image

When no location is selected, an add location option is available:
image

Create Option in Dialog:
image

valonm:
TODO UI Issues:
- When Creating a new mission, location autocomplete is populated with a shell location, which it thinks is editable. It should really just be undefined and the option to create a territory should appear.
- When editing a location the toast message says it was created successfully -- need to either handle more use cases with the toast, or just eliminate it for now.

valonm:
Added Driving Directions Componentry

image
image
image

src/components/ConfirmDialog.tsx
```diff
@@ -0,0 +1,19 @@
...
+    onConfirm();
+    onClose();
+  };
+  return (
+    <Dialog open={open} onClose={onClose}>
+      <DialogContent>{prompt}</DialogContent>
+      <DialogActions>
+        <Button onClick={onClose}>Cancel</Button>
+        <Button variant="contained" onClick={handleConfirm}>
+          confirm
```
mcosand, line 14: casing. Is "OK" more common?

mcosand (approved):
I don't currently have opinions on the flow of locations in the edit event dialog. It seems there are weird cases where a user edits a location to  fix an error (typo in an address) vs editing a location that's actually a new location (CP moved) when both old and new are likely never to be used again, etc. Assume Valon has thought about these and we can iterate if needed.

src/components/ConfirmDialog.tsx
valonm: Button converts it to all uppercase. I am aligned on "OK", will swap that out.

valonm:
> I don't currently have opinions on the flow of locations in the edit event dialog. It seems there are weird cases where a user edits a location to fix an error (typo in an address) vs editing a location that's actually a new location (CP moved) when both old and new are likely never to be used again, etc. Assume Valon has thought about these and we can iterate if needed.

I don't anticipate that locations will be edited frequently. But when they are edited from the Activity page they only change the location data for that activity, except in the case that the user selects the "Update Saved Location" option. If this becomes problematic for some reason, we can always remove the "Update Saved Location" feature and only allow locations to be edited from the admin page.

Another functionality I had in mind but have not implemented was that there may be use cases where we want to updated the saved location, then pull that new information into an existing activity location. Like, "Sync from Saved Location" or something to that effect. But, again, I expect this to be an edge case. I wasn't planning to introduce that complexity unless there was a good value prop. I think this would become more of a necessity if Saved Locations cannot be updated edited from the activity because I'd want to save the change, then pull the update into the mission.

There is technically a workaround for this: if you change the location selection to a different selection, then back, it will update the Activity Location to the latest Saved Location data. It does not preserve the state of the Activity Location if it is unselected -- this would only be persisted if the activity is subsequently saved. i.e. if you change the location and lose temporary edits you can cancel the activity edit to restore the activity (and location) to it's previous state.

PR #82, opened 2024-01-28
Fix preferences race condition

valonm:
I found a minor bug with the Preferences dialog: On first open, the dialog loads with the default state. If you cancel and re-open it, it has the actual saved state. There seems to be a race condition between the dialog initializing and the redux store loading. I refactored the AppMenu so that the Preferences Dialog is only rendered if it is open; this means that the dialog does not check the store for the user's preferences until it is accessed.

### DIFF
--- a/src/components/AppMenu.tsx
+++ b/src/components/AppMenu.tsx
@@ -68,7 +68,7 @@ export function AppMenu() {
           </MenuItem>
         </Menu>
       </div>
-      {openPreferences && <PreferencesDialog open={openPreferences} onClose={handleClosePreferences} />}
+      <PreferencesDialog open={openPreferences} onClose={handleClosePreferences} />
     </>
   );
 }
--- a/src/components/AppMenu.tsx
+++ b/src/components/AppMenu.tsx
@@ -68,7 +68,7 @@ export function AppMenu() {
           </MenuItem>
         </Menu>
       </div>
-      <PreferencesDialog open={openPreferences} onClose={handleClosePreferences} />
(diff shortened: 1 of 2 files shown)


valonm:
[Fix Preferences Race Condition](https://trello.com/c/aun5OkVK/71-fix-preferences-race-condition)

src/components/AppMenu.tsx
```diff
@@ -68,7 +68,7 @@ export function AppMenu() {
           </MenuItem>
         </Menu>
       </div>
-      <PreferencesDialog open={openPreferences} onClose={handleClosePreferences} />
+      {openPreferences && <PreferencesDialog open={openPreferences} onClose={handleClosePreferences} />}
```
colincwilliams-kcsara, line 71: Given it's not going to be loaded at all if openPreferences is false, does the "open" property need to exist on PreferencesDialog anymore?

valonm: `open` is a required prop on Dialog

valonm:
Ok, I went a completely different route to handle the issue more-directly. Basically, the form shouldn't initialize until it is rendered; but it was being rendered immediately (albeit hidden) because the dialog renders hidden on page load. Instead of expecting the caller to conditionally render, the dialog, I have separated the form to it's own component so that the dialog can conditionally render it directly.

While I was tinkering, I also updated the form to useForm which will make it a lot easier to add fields later.

colincwilliams-kcsara (approved):
Looks great!

PR #83, opened 2024-02-02
Roster Print/Save PDF View and CSV Download

valonm:
Refactor of the Roster page to implement cleaner print view and add option for CSV download.
* The print/save PDF view only renders the table; auto pops the print dialog; redirects back to the roster view after the print dialog closes.
* Download to CSV was a feedback ask after the initial roster view rollout.

image

image

image

### DIFF
--- a/src/components/activities/RosterView.tsx
+++ b/src/components/activities/RosterView.tsx
@@ -1,6 +1,7 @@
 import { Box, Button, Divider, Paper, Stack, Table, TableBody, TableCell, TableHead, TableRow, Typography } from '@mui/material';
 import { differenceInCalendarDays, format as formatDate } from 'date-fns';
-import { useRouter } from 'next/navigation';
+import { forwardRef, useRef } from 'react';
+import { useReactToPrint } from 'react-to-print';
 
 import { ToolbarPage } from '@respond/components/ToolbarPage';
 import { useAppSelector } from '@respond/lib/client/store';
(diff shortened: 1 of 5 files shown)


src/components/activities/RosterView.tsx
```diff
@@ -1,102 +1,109 @@
...
 }
 
+export function RosterPrint({ activityId }: { activityId: string }) {
+  const router = useRouter();
+  const activity = useAppSelector(buildActivitySelector(activityId));
+  const rosterEntries = getRosterEntries(activity);
+  window.onafterprint = () => router.back();
+  setTimeout(() => {
+    window.print();
+  }, 1000);
```
colincwilliams-kcsara, line 65: Why wait a full second? Without having used the feature, I'm not exactly sure how it feels, but in my mind that's a pretty long time and would surprise a user. Basically they land on the page, wonder what is happening, and then the print dialog magically appears.

Is that just ensuring the page has fully rendered before popping the dialog?

colincwilliams-kcsara (approved):
Looks good, just the one question and I'm good with it going in as-is assuming there's a good reason for the full 1 second delay.

src/components/activities/RosterView.tsx
valonm: Yeah, there might be a better way to do it, but I was having a problem where the print call happens before the component actually renders. I couldn't find a better solution; open to suggestions.

valonm: [react-to-print](https://www.npmjs.com/package/react-to-print) looks like an option but I was trying to avoid a dependency.

PR #87, opened 2024-03-15
Participant ETA

valonm:
Participants in "Standby" or "SignedIn" status will see `ParticipantEtaUpdater`. If the user reports an ETA, then it will be displayed next to their current status in the roster tile.

- When the Participant *has not* reported an ETA they will see options to quickly report a 15, 30, or 60 minute ETA, or to enter a specific time. Clicking on the "clock" button opens an inline `DateTimePicker`.

- When the Participant *has* reported an ETA it will show the current value and has options to increment by 1 minute or to clear the value.

- If the Participant is not in "Standby" or "SignedIn" status the `ParticipantEtaUpdater` is hidden and the roster tile will not display the ETA.

## Screenshots
Mobile: iPhone 5/SE

image

image

image

image

image

image

### DIFF
--- a/src/components/InlineTimeEdit.tsx
+++ b/src/components/InlineTimeEdit.tsx
@@ -1,0 +1,16 @@
+import { Close } from '@mui/icons-material';
+import { DateTimePicker } from '@mui/x-date-pickers';
+import { DateOrTimeViewWithMeridiem } from '@mui/x-date-pickers/internals/models';
+
+import { IconButton, Stack } from './Material';
+
+export function InlineTimeEdit({ label, format, openTo, onChange, onClose }: { label: string; format: string; openTo: DateOrTimeViewWithMeridiem | undefined; onChange: (time: number | null) => void; onClose: () => void }) {
+  return (
+    <Stack flexGrow={1} direction="row" justifyContent="space-between">
+      <DateTimePicker value={new Date().getTime()} label={label} format={format} onAccept={onChange} onClose={onClose} openTo={openTo} />
+      <IconButton disableRipple onClick={onClose}>
+        <Close />
+      </IconButton>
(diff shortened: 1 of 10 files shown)


valonm:
[Add Field for ETA to CP](https://trello.com/c/0F66jJKF/15-add-field-for-eta-to-cp)

valonm:
Added a switch to sort responders either alphabetically or by ETA.

Previously responders were not sorted; the array is just in the order of that participants were added.

## Sort Behavior
**Name:** Sorts the list by First Name.
**ETA:** Sorts the list by status then ETA with the following priority:
1. Participants in Standby or SignedIn with an ETA; Sorted by ETA ascending
2. Participants in Standby or SignedIn with NO ETA
3. Participants in other Statuses

image
image

mcosand:
There are some elements I'd like to work on here. I don't know if it's better as a github thread, Slack conversation, or a meeting.
- Should ETA take the bottom right corner of the card, instead of parentheticals?
- I like the idea of the ETA control being prominent so it's easy to update, but it's prominent in areas of the app where I really want to maximize real estate (like the roster). Thinking of ways we can manage this.
- When I finally get "sign in other user", I'm thinking about moving the status updater to a floating action button and full screen page/dialog. ie: I'm thinking about UI changes for the status updater and don't want to paint ourselves into a corner.
- I think sort is more complicated than a switch (I'd also like to sort by status and last modified). Still trying to figure out how to do that in limited real estate.
- (minor) I don't care about 1-minute updates to ETA. I'd start at +/-10min or 15min updates.

valonm:
Per discussion with Matt, I'm going to merge as-is but the UI/UX for this feature deserves some more iteration.

PR #88, opened 2024-03-19
Participant Edit Miles

valonm:
This commit enables editing of participant mileage, separate from a "Sign Out" status update.
- re-implemented the participant miles input so that it is re-usable
- re-arranged the participant dialog a bit so the edit functionality has enough space and the UI is not too small to tap on in Mobile
- added edit miles functionality in the participant dialog.

## Refactored Miles Input on the Sign Out Form
image
image

## Rearranged Participant Dialog
I changed it to 2 columns, instead of three on desktop. It was always 1 column on mobile (no change.) I also made the Total Hours and Total Miles a larger to match the timeline font. The Total Miles row can be clicked/tapped to edit, just like the existing timeline edit functionality.

image
image

## Editing on the Dialog
image
image

### DIFF
--- a/src/components/ParticipantMilesInput.tsx
+++ b/src/components/ParticipantMilesInput.tsx
@@ -13,19 +13,20 @@ type MilesAction = {
 };
 
 const reducer: Reducer<ParticipantMilesState, MilesAction> = (state: ParticipantMilesState, action: MilesAction) => {
+  const hasMiles = action.miles > 0;
   switch (action.type) {
     case 'set_total': {
-      return { ...state, miles: action.miles };
+      return { ...state, miles: hasMiles ? action.miles : state.initialMiles };
     }
     case 'add_miles': {
-      return { ...state, miles: state.initialMiles + action.miles };
+      return { ...state, miles: state.initialMiles + (hasMiles ? action.miles : 0) };
     }
   }
 };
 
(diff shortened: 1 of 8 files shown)


valonm:
[Edit Mileage](https://trello.com/c/xtHIjXtu/73-edit-mileage)

mcosand (commented):
The action/reducer looks OK to me.
I don't have strong feelings on the rest. This does mean we lose what form validation we had for the miles field? Don't know that it's worth worrying about keeping, but I was hoping we'd be able to have clean validation/error messages that kept the database in a good state. You may have been playing the cards I dealt you.

valonm:
> This does mean we lose what form validation we had for the miles field?

The only validation that the status updater is doing is that the miles have to be non-negative. `ParticipantMilesUpdater` is enforcing that rule, although, it doesn't present it as an error; it just doesn't enable the submit button when there is invalid input.

valonm:
Fixed a minor bug I found while looking into the validation question. The `ParticipantMiles` component on the `RosterPanel` maintains a miles state, but was not passing updates along with the participant.

valonm:
I have resolved the issue that Matt pointed out, regarding the parent not being able to handle validation (e.g. `useForm` resolvers) although, I have also built it to prevent negative value input.

I also did some (probably unnecessary) refactoring to `useReducer` which reduced the complexity of the internal component state, but required a bunch of typescript boilerplate -- it was probably a wash, but I learned some stuff about `useReducer`; so that's good :-)

PR #89, opened 2024-03-23
Phone and Email on Participant Dialog

valonm:
This commit updates the Participant Dialog to include the Participant's `mobilephone` and `email` attributes. These values are fetched from the member provider on demand (on load) and are not persisted locally or in the app database.

image

### DIFF
--- a/src/components/activities/RosterPanel.tsx
+++ b/src/components/activities/RosterPanel.tsx
@@ -6,6 +6,8 @@ import { useTheme } from '@mui/material/styles';
 import { FunctionComponent, ReactNode, useEffect, useState } from 'react';
 
 import { Box, Dialog, DialogContent, DialogTitle, Paper, Stack, Typography, useMediaQuery } from '@respond/components/Material';
+import { apiFetch } from '@respond/lib/api';
+import { MemberInfo } from '@respond/lib/server/memberProviders/memberProvider';
 import { Activity, getOrganizationName, getStatusCssColor, getStatusText, isActive, Participant, ParticipantStatus, ParticipantUpdate, ParticipatingOrg } from '@respond/types/activity';
 
 import ParticipantTimeline from './ParticipantTimeline';
@@ -17,6 +19,20 @@ interface RosterPanelProps {
   participantRowComponent: FunctionComponent<{ orgs: Record<string, ParticipatingOrg>; participant: Participant; onClick?: () => void }>;
   onClick?: (participant: Participant) => void;
 }
+
+const findMember = async (orgId: string, memberId: string) => {
+  return (await apiFetch<{ data: MemberInfo }>(`/api/v1/organizations/${orgId}/members/${memberId}`)).data;
+};
+
+const formatPhoneNumber = (phoneNumberString: string, trimIntlCode?: boolean) => {
+  const cleaned = ('' + phoneNumberString).replace(/\D/g, '');
+  const match = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/);
+  if (match) {
+    const intlCode = match[1] ? '+1 ' : '';
+    return [trimIntlCode ? '' : intlCode, '(', match[2], ') ', match[3], '-', match[4]].join('');
+  }
+  return null;
+};
 
 export function RosterPanel({ activity, filter, participantContainerComponent: Participants, participantRowComponent: Participant, onClick }: RosterPanelProps) {
   let participants = Object.values(activity.participants);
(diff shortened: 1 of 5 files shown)


valonm:
[Additional Member Info on Participant Card - Email, Phone # (maybe other info?)](https://trello.com/c/xNC0qe3u/66-additional-member-info-on-participant-card-email-phone-maybe-other-info)

src/lib/server/memberProviders/d4hMembersProvider.ts
```diff
@@ -78,6 +78,27 @@ export default class D4HMembersProvider implements MemberProvider {
     }
   }
 
+  async findMember(memberId: string) {
+    await this.initialize();
+    for (const token in this.tokenFetchInfo) {
```
mcosand, line 83: Right now, every unit is using the same D4H token, so this might only be N=1, but it's possible for each unit to have their own token, where we'd be making N=(number of units) network requests, hoping that one of the tokens had permissions to complete the request. I think this method should look more like the top of the `getMemberPhoto` method - we should already have the member data cached in the tokenFetchInfo object. `const member = this.tokenFetchInfo[token].lookup[memberId].response`

src/components/activities/RosterPanel.tsx
```diff
@@ -18,6 +20,20 @@ interface RosterPanelProps {
   onClick?: (participant: Participant) => void;
 }
 
+const findMember = async (orgId: string, memberId: string) => {
+  return (await apiFetch<{ data: MemberInfo }>(`/api/v1/organizations/${orgId}/members/${memberId}`)).data;
+};
+
+const formatPhoneNumber = (phoneNumberString: string, trimIntlCode?: boolean) => {
```
mcosand, line 27: Do we need `trimIntlCode`? If we need it, can we give it a default value so most callers can ignore it?

src/components/activities/RosterPanel.tsx
```diff
@@ -58,6 +74,17 @@ export function RosterRowCard({ status, children, onClick, ...props }: PaperProp
...
+  const [memberInfo, setMemberInfo] = useState<MemberInfo | undefined>();
+
+  useEffect(() => {
+    if (participant) getMemberInfo(participant);
+  }, [participant]);
+
+  const getMemberInfo = async (participant: Participant) => {
+    const member = await findMember(participant.organizationId, participant.id);
+    setMemberInfo(member);
+    console.log('dialog', member);
```
mcosand, line 86: debug cruft

src/components/activities/RosterPanel.tsx
```diff
@@ -6,6 +6,8 @@ import { useTheme } from '@mui/material/styles';
 import { FunctionComponent, ReactNode, useEffect, useState } from 'react';
 
 import { Box, Dialog, DialogContent, DialogTitle, Paper, Stack, Typography, useMediaQuery } from '@respond/components/Material';
+import { apiFetch } from '@respond/lib/api';
+import { MemberInfo } from '@respond/lib/server/memberProviders/memberProvider';
```
mcosand, line 10: Client side code is importing server side classes. This should be avoided. One reason is that the server-side class may have sensitive information that is not intended to be exposed to any client. `/lib/types` would be a better place for the class/interface.

mcosand (changes requested):

src/components/activities/RosterPanel.tsx
valonm: removed

src/components/activities/RosterPanel.tsx
valonm: refactored to explicitly default false

src/components/activities/RosterPanel.tsx
valonm: created a separate client side interface

src/lib/server/memberProviders/d4hMembersProvider.ts
valonm: refactored

src/lib/server/memberProviders/d4hMembersProvider.ts
```diff
@@ -78,6 +78,18 @@ export default class D4HMembersProvider implements MemberProvider {
...
 
+  async getParticipantInfo(memberId: string) {
+    await this.initialize();
+    for (const token in this.tokenFetchInfo) {
+      const member = this.tokenFetchInfo[token].lookup[memberId].response;
+      if (!member) continue;
+      return {
+        email: member.email,
+        mobilephone: member.mobilephone,
+      } as MemberInfo;
```
mcosand, line 89: This should be `ParticipantInfo`? One thing I've done in the past is assign to a variable first, to make sure the type is correct, and then return:
```suggestion
      const result: ParticipantInfo {
        email: member.email,
        mobilephone: member.mobilephone,
      };
      return result;
```

mcosand (approved):
One type thing to take care of, otherwise looks pretty good.

PR #94, opened 2025-02-09
add base kiosk mode

valonm:
This enhancement adds a new "Base Kiosk" mode, accessible from Desktop Activity pages.

image

Accessing this mode allows users to manage their roster entries from a shared computer; e.g. at the command post. Similarly, it will allow users to add/update other responders to the roster.

On the Kiosk page, start by selecting which Organization to associate new roster entries with:

image

Next, a search box appears which queries the MemberProvider for the selected Organization, entering a few characters will start a query:

image

The status updater on this page updates the status of the currently selected member:

image

### DIFF
--- a/src/app/api/v1/organizations/route.ts
+++ b/src/app/api/v1/organizations/route.ts
@@ -1,0 +1,15 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import * as Mongo from '@respond/lib/server/mongodb';
+
+export async function GET(_request: NextRequest) {
+  const organizations = await Mongo.getOrganizations();
+  if (!organizations) {
+    return NextResponse.json({ status: 'get organizations failed' }, { status: 500 });
+  }
+  console.log(organizations);
+  return NextResponse.json({
+    status: 'ok',
+    data: organizations,
+  });
+}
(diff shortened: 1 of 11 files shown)


src/app/api/v1/organizations/route.ts
```diff
@@ -0,0 +1,15 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import * as Mongo from '@respond/lib/server/mongodb';
+
+export async function GET(_request: NextRequest) {
+  const organizations = await Mongo.getOrganizations();
+  if (!organizations) {
+    return NextResponse.json({ status: 'get organizations failed' }, { status: 500 });
+  }
+  console.log(organizations);
```
valonm, line 10: need to remove this

valonm:
I'm noticing strange behavior of the timeline component in the context of the kiosk page; it seems to apply the wrong time on new entries, however the correct time is recorded in the database. The correct times appear on page refresh. I'll need to look into it some more.

valonm:
I've resolved the state issue on the kiosk page, but now I have a similar problem on the participant dialog.

The timeline in the ParticipantDialog does not re-render when the participant status is updated. I think the issue is related to useState storing a shallow copy of the participants array, however it looks like that should be updated by the useEffect that watches activity, filter and sortEta.

valonm:
Added status updater on Participant dialog so that In Towns, Base (or other users in general) can update any responders status.

image

PR #100, opened 2025-02-22
Enhancements to the ActivityList Page

valonm:
Updates to the ActivityList page for visual consistency and to add pagination and filtering.

image


mcosand (approved):
cursory review. I had concerns earlier over the use of the data grid and its row limit, but I think the filter function alleviates that, especially for an admin/less often accessed part of the site.

PR #156, opened 2026-05-29
Fix: Cache exceeding local storage quota causing stale UI

### LINKED ISSUE #152: Bug: Offline Cache not Updating
When I open the respond app I often see missions and events from months ago.
It takes seconds until the new information is streamed in.

See video below (test data on "Fast 4G" connection):

https://github.com/user-attachments/assets/c8316ebe-64b5-45d1-a01b-63ec7e3c80de

### LINKED ISSUE #153: Bug: Error message shown before activity is loaded
When opening a mission or event activity, an error message is shown before the activity is loaded, which can take several seconds.
This can be miss-leading and confusing to users, as it was to me the first time I saw this.

Video simulating Chrome "Fast 4G" network with test data.

https://github.com/user-attachments/assets/a82ead92-d182-4e24-8b0f-d239ccedbc10

kohlikohl:
The PR fixes #153 and #152 by doing two things:
* filtering the activities on the server to reduce payload as well as on the client to reduce localStorage size
* replace the localStorage cache with the filtered items on update and reload actions

### Filter logic
Only keep items that
* are active, which is asserted by checking if an item as an `endTime`
* if they are activities that are in the future
* if they are activities that have either started or ended within the last 90 days

### Validation
* The changes were validated locally with seed data exceeding local storage quota as well as exceeding events spanning back 2 years. 
* The bugs were reproducible in the local setup and are fixed with this change

### DIFF
--- a/src/lib/state/activityVisibility.ts
+++ b/src/lib/state/activityVisibility.ts
@@ -1,0 +1,28 @@
+import type { Activity } from '@respond/types/activity';
+
+/**
+ * Filter activities by
+ * * Active activities (no end time) are always included.
+ * * Future activities (start time in the future) are always included.
+ * * Completed activities are included only when their latest activity time is inside the retention window.
+ *
+ * This is used to determine which activities should be included in the initial state when the app loads, and also to 
+ * filter activities before saving them to localStorage for caching purposes.
+ */
+export const INITIAL_ACTIVITY_HISTORY_DAYS = 90;
+export const INITIAL_ACTIVITY_HISTORY_MS = INITIAL_ACTIVITY_HISTORY_DAYS * 24 * 60 * 60 * 1000;
+
+export function isActivityIncludedInInitialState(activity: Activity, now = Date.now(), historyMs = INITIAL_ACTIVITY_HISTORY_MS) {
+  // Active activities must remain in the initial cache even if they started long ago.
+  if (!activity.endTime) return true;
+  // Future activities must remain visible before they begin.
+  if (activity.startTime > now) return true;
+
+  // Completed activities are included only when their latest activity time is inside the retention window.
+  const mostRecentActivityTime = Math.max(activity.startTime, activity.endTime);
+  return mostRecentActivityTime >= now - historyMs;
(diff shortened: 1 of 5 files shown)


src/lib/state/activityVisibility.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function isActivityIncludedInInitialState(activity: Activity, now = Date.now(), historyMs = INITIAL_ACTIVITY_HISTORY_MS) {
+  // Active activities must remain in the initial cache even if they started long ago.
+  if (!activity.endTime) return true;
+  // Future activities must remain visible before they begin.
+  if (activity.startTime > now) return true;
+
+  // Completed activities are included only when their latest activity time is inside the retention window.
+  const mostRecentActivityTime = Math.max(activity.startTime, activity.endTime);
+  return mostRecentActivityTime >= now - historyMs;
```
valonm, line 23: End date should always be the larger value; this can just evaluate `activity.endTime >= now - historyMs`

valonm (commented):
This change causes some second-order effects (which may also already be present due to the standing issue); missions/events that are not in the cache do not show up in the history, e.g. on the `/mission` or `/event` paths; nor can they be referenced directly, e.g. `/mission/<id>`. They essentially don't exist unless they've been persisted to local storage.

Features that depend on the cache containing the full history will also need to be reimplemented. I'm open to discussion on whether that is a blocker for this as a quick fix. Users accessing activities older than 90 days is probably an edge case.

src/lib/state/activityVisibility.ts
valonm: on second thought, I'd have to double check how we handle events that are closed before the `startTime`.

kohlikohl: Seem like the end time is recorded as is even if it is before the start date.
It might not matter if we pick the higher of the two values or always stick to `endTime` as the difference in what is displayed would be minimal. 
Happy to change it to `activity.endTime >= now - historyMs`.

PR #160, opened 2026-06-02
Fix: Cache exceeding local storage quota causing stale UI

### LINKED ISSUE #152: Bug: Offline Cache not Updating

### LINKED ISSUE #153: Bug: Error message shown before activity is loaded

kohlikohl:
The PR fixes #153 and #152 by doing two things:
* filtering the activities on the server to reduce payload as well as on the client to reduce localStorage size
* replace the localStorage cache with the filtered items on update and reload actions
* The change breaks the `/mission` and `/event` routes which are supposed to show all known activities. To fix this:
  * Changed the `ActivityListPage` to fetch the data from `/api/v1/${activityType}` API, which is serving *all* activities
  * Introduced `src/app/api/v1/(activities)/activities/[activityId]/route.ts` API route to fetch specific activity so that activities that are not cached in `localStorage` can still be viewed

### Filter logic
Only keep items that
* are active, which is asserted by checking if an item as an `endTime`
* if they are activities that are in the future
* if they are activities that have either started or ended within the last 90 days

### Validation
* The changes were validated locally with seed data exceeding local storage quota as well as exceeding events spanning back 2 years. 
* The bugs were reproducible in the local setup and are fixed with this change

### DIFF
--- a/src/app/api/v1/(activities)/activities/[activityId]/route.ts
+++ b/src/app/api/v1/(activities)/activities/[activityId]/route.ts
@@ -1,0 +1,18 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import { getCookieAuth, userFromAuth } from '@respond/lib/server/auth';
+import { getServices } from '@respond/lib/server/services';
+
+export async function GET(_request: NextRequest, { params }: { params: { activityId: string } }) {
+  const user = userFromAuth(await getCookieAuth());
+  if (user == null) {
+    return NextResponse.json({ status: 'not authenticated' }, { status: 401 });
+  }
+
+  const activity = (await (await getServices()).stateManager.getAllActivities()).find((a) => a.id === params.activityId);
+  if (!activity) {
+    return NextResponse.json({ status: 'not found' }, { status: 404 });
+  }
+
+  return NextResponse.json({ status: 'ok', data: activity });
+}
--- a/src/components/activities/ActivityListPage.tsx
+++ b/src/components/activities/ActivityListPage.tsx
@@ -1,10 +1,11 @@
 import { Breadcrumbs, Paper, Typography } from '@mui/material';
 import { DataGrid, GridColDef, GridToolbarContainer, GridToolbarFilterButton } from '@mui/x-data-grid';
 import Link from 'next/link';
+import { useEffect, useState } from 'react';
 
 import { ToolbarPage } from '@respond/components/ToolbarPage';
-import { useAppSelector } from '@respond/lib/client/store';
-import { buildActivityTypeSelector, getActivityPath } from '@respond/lib/client/store/activities';
+import { apiFetch } from '@respond/lib/api';
+import { getActivityPath } from '@respond/lib/client/store/activities';
 import { Activity, ActivityType } from '@respond/types/activity';
 
 import { RelativeTimeText } from '../RelativeTimeText';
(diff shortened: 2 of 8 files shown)


src/components/activities/ActivityListPage.tsx
```diff
@@ -30,15 +31,62 @@ const columns: GridColDef[] = [
...
+      loading={loading}
+      initialState={{ pagination: { paginationModel } }}
+      pageSizeOptions={[15, 25, 50]}
+      density="compact"
+      autoHeight
+      slots={{ toolbar: GridToolbar }}
+      sx={
+        loading
+          ? {
+              '--unstable_DataGrid-overlayBackground': (theme) => theme.palette.background.paper,
```
kohlikohl, line 48: This change improves the loading spinner, so that it is not showing over a red background making it look like an error. Instead when the list is loading, it shows the spinner on white background.

Image

src/components/activities/ActivityListPage.tsx
```diff
@@ -30,15 +31,62 @@ const columns: GridColDef[] = [
...
 const paginationModel = { page: 0, pageSize: 15 };
 
-function ActivityList({ activities }: { activities: Activity[] }) {
-  return <DataGrid rows={activities} columns={columns} initialState={{ pagination: { paginationModel } }} pageSizeOptions={[15, 25, 50]} density="compact" autoHeight slots={{ toolbar: GridToolbar }} />;
+function ActivityList({ activities, loading }: { activities: Activity[]; loading: boolean }) {
+  return (
+    <DataGrid
+      rows={activities}
+      columns={columns}
+      loading={loading}
```
kohlikohl, line 39: This adds a loading spinner while the data is fetched from the API instead of a red row with the message "no rows".

src/app/api/v1/(activities)/activities/[activityId]/route.ts
```diff
@@ -0,0 +1,18 @@
+import { NextRequest, NextResponse } from 'next/server';
```
Comment (kohlikohl) [inline src/app/api/v1/(activities)/activities/[activityId]/route.ts:1]: New route to enable loading of activities that are not cached in the 90 day cache.

src/components/activities/ActivityPage.tsx
```diff
@@ -17,18 +19,40 @@ import { MobileActivityPage } from './MobileActivityPage';
...
+    document.title = `${displayActivity?.idNumber} ${displayActivity?.title}`;
+  }, [displayActivity?.idNumber, displayActivity?.title]);
+
+  useEffect(() => {
+    if (activity) return;
+
+    let cancelled = false;
+    setFetchedActivity(undefined);
+
+    apiFetch<{ data?: Activity }>(`/api/v1/activities/${activityId}`)
```
kohlikohl, line 35: When `activity` is not set it means that it doesn't exist in localStorage likely because it is more than 90 days old. Therefore we need to fetch it from the server.

PR #161, opened 2026-06-03
Show empty roster rows

kohlikohl:
This PR fixes issue #159.

### Changes
* pads roster with 20 empty rows when a mission does not have any participants yet
* as participants are signing in, the roster is filled in and if the 20 rows are exceeded additional rows are created
* showing 20 empty/partially filled rows to allow for printing a whole portrait sheet of rows.

### Verification
* local with test data

**Mission Roster with no participants**
image

**Mission Roster with one participant**
image

**Mission Roster with more than 20 participants**
image

### DIFF
--- a/src/components/activities/RosterView.tsx
+++ b/src/components/activities/RosterView.tsx
@@ -20,6 +20,8 @@ interface RosterEntryProps {
 }
 
 const headerCellStyle = { fontWeight: 700, width: 20 };
+const rosterRowStyle = { height: 45 };
+const MIN_ROSTER_ROWS = 5;
 const columnBordersStyle = {
   [`& .${tableCellClasses.root}`]: {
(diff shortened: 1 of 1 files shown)


src/components/activities/RosterView.tsx
```diff
@@ -20,6 +20,8 @@ interface RosterEntryProps {
 }
 
 const headerCellStyle = { fontWeight: 700, width: 20 };
+const rosterRowStyle = { height: 45 };
+const MIN_ROSTER_ROWS = 5;
```
valonm: The print view includes the blank rows. I think we should either a.) drop the empty rows when printing, or b.) have a minimum of 20 rows, which fills an 8.5x11 page, which would be more useful for activities where we want to include folks that are signed up already but have extra rows to write in. My preference would be the latter.

Image

kohlikohl: Sounds good! Updated it to 20 rows.

PR #162, opened 2026-06-05
Add API for external integrations (SMR Member App Project)

NickConstantine:
# Summary

This pull request adds a small, self-contained API that lets an external application read mission and event data and push participant status updates back into Respond over HTTP – without using the interactive (cookie / Google OAuth) session the web UI relies on. This would allow support for SMR's Member's App to:

- Read the missions and events and response timelines its members are involved in
- Write back its own members' status (signed in, responding, etc.) so Respond stays the source of truth

Today there is no non-interactive way for an external app to do this safely. This PR adds one.

# How it works

- **Authentication:** SMR's member app sends an `x-api-key` header. Keys are configured via a new `PARTNER_API_KEYS` environment variable (a JSON map of key → orgId). Each key is bound to exactly one organization.
- **Scoping:** every request is locked to the key's organization (and its configured partner organization team). A partner can never read or write data outside its own org.
- **Writes go through the normal pipeline:** participant updates are dispatched via `handleIncomingAction()` – exactly like the web UI – so in-memory state, MongoDB, and connected websocket clients all stay in sync. No raw DB writes.

# Endpoints (all under `/api/partner/v1`)

| Method | Path | Purpose |
|---|---|---|
| GET | `/activities` | List in-scope missions/events (optional `?type=missions\|events`) |
| GET | `/activities/{id}` | Get a single in-scope activity |
| POST | `/activities/{id}/participants` | Write back a participant status update |

# Safety guarantees

- **Org isolation –** reads are filtered to the partner's team org; out-of-scope activities return 404.
- **Write guardrail –** a partner org team may only write participants coded to its own org team. The participant's `organizationId` is forced to the partner's org, and any attempt to modify a participant coded to a different org is rejected with 403.
- **Timing-safe key comparison –** keys are compared with `crypto.timingSafeEqual` to avoid timing attacks.
- **Validation –** malformed bodies and invalid status values return 400.

# Testing

- 12 unit tests covering auth and org-scoping logic.
- 11 end-to-end integration tests that drive the real route handlers against an in-memory data layer using the actual activity reducer – proving reads, scoped filtering, successful SMR write-back (visible on re-read), and the 403 cross-org guardrail.
- **Full suite: 26/26 passing.** New code passes lint.

# Configuration

Add to the server environment (kept out of source / the repo):

```
PARTNER_API_KEYS={"<random-secret-key>":"<partner-org-id>"}
```

Generate a key with e.g. `openssl rand -hex 32`, then share it with the partner over a secure channel (password vault or one-time secret link).

# Files changed (9 files, +592 / −1)

- `src/lib/server/partnerAuth.ts` – API key authentication
- `src/lib/server/partnerScope.ts` – org-scoping helpers
- `src/app/api/partner/v1/activities/route.ts` – list endpoint
- `src/app/api/partner/v1/activities/[activityId]/route.ts` – single-activity endpoint
- `src/app/api/partner/v1/activities/[activityId]/participants/route.ts` – write-back endpoint
- Three test files – unit + integration tests
- `jest.config.mjs` – fixed a latent `@respond/*` module-mapping bug so tests resolve (repo had no tests previously)

### DIFF
--- a/src/app/api/partner/v1/activities/[activityId]/participants/route.ts
+++ b/src/app/api/partner/v1/activities/[activityId]/participants/route.ts
@@ -1,0 +1,95 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+import { authenticatePartner } from '@respond/lib/server/partnerAuth';
+import { getPartnerOrgScope, isActivityInScope } from '@respond/lib/server/partnerScope';
+import { getServices } from '@respond/lib/server/services';
+import { ActivityActions } from '@respond/lib/state/activityActions';
+import { ParticipantStatus } from '@respond/types/activity';
+
+interface ParticipantUpdateBody {
+  participantId?: string;
+  firstname?: string;
+  lastname?: string;
+  status?: number;
+  time?: number;
+  miles?: number;
+  eta?: number;
+}
+
+const VALID_STATUSES = new Set<number>(Object.values(ParticipantStatus).filter((v): v is number => typeof v === 'number'));
+
+/**
+ * POST /api/partner/v1/activities/[activityId]/participants
+ *
+ * Pushes a participant status update back into Respond on behalf of the partner
+ * organization. The update is dispatched through the normal action pipeline
+ * (handleIncomingAction) so in-memory state, MongoDB, and all connected
+ * websocket clients stay in sync — it does NOT write to MongoDB directly.
+ *
+ * SCOPE GUARANTEE: a partner may only write participant records coded to its
+ * own organization. The participant's organizationId is forced to the partner's
+ * org id, and any attempt to modify a participant already coded to a different
+ * org is rejected with 403.
+ *
+ * Auth: `x-api-key` header.
+ *
+ * Body:
+ *   { participantId, firstname?, lastname?, status, time?, miles?, eta? }
+ */
+export async function POST(request: NextRequest, { params }: { params: { activityId: string } }) {
+  const partner = authenticatePartner(request);
+  if (!partner) {
+    return NextResponse.json({ status: 'unauthorized' }, { status: 401 });
+  }
+
+  let body: ParticipantUpdateBody;
+  try {
+    body = await request.json();
+  } catch {
+    return NextResponse.json({ status: 'bad request', message: 'Invalid JSON body' }, { status: 400 });
+  }
+
+  const { participantId, firstname, lastname, status, time, miles, eta } = body;
+
+  if (!participantId || typeof participantId !== 'string') {
+    return NextResponse.json({ status: 'bad request', message: 'participantId is required' }, { status: 400 });
+  }
+  if (typeof status !== 'number' || !VALID_STATUSES.has(status)) {
+    return NextResponse.json({ status: 'bad request', message: 'status must be a valid ParticipantStatus value' }, { status: 400 });
+  }
+
+  const services = await getServices();
+  const all = await services.stateManager.getAllActivities();
+  const activity = all.find((a) => a.id === params.activityId);
+
+  const scope = await getPartnerOrgScope(partner);
+  if (!activity || !isActivityInScope(activity, scope)) {
+    return NextResponse.json({ status: 'not found' }, { status: 404 });
+  }
+
+  // SMR scope enforcement: never allow writing over a participant that belongs
+  // to another organization.
+  const existing = activity.participants?.[participantId];
+  if (existing && existing.organizationId !== partner.orgId) {
+    return NextResponse.json({ status: 'forbidden', message: 'Participant is coded to a different organization' }, { status: 403 });
+  }
+
+  const action = ActivityActions.participantUpdate(
+    params.activityId,
+    participantId,
+    firstname ?? existing?.firstname ?? '',
+    lastname ?? existing?.lastname ?? '',
+    partner.orgId, // forced to the partner's org — cannot write for other orgs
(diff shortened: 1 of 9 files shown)


src/app/api/partner/v1/activities/[activityId]/participants/route.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  let body: ParticipantUpdateBody;
+  try {
+    body = await request.json();
+  } catch {
+    return NextResponse.json({ status: 'bad request', message: 'Invalid JSON body' }, { status: 400 });
+  }
+
+  const { participantId, firstname, lastname, status, time, miles, eta } = body;
```
Comment (kohlikohl) [inline src/app/api/partner/v1/activities/[activityId]/participants/route.ts:57]: Where does the `participantId` come from? Does he external SMR app resolve to the same ids as respond does?

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
```diff
@@ -0,0 +1,95 @@
...
+  // to another organization.
+  const existing = activity.participants?.[participantId];
+  if (existing && existing.organizationId !== partner.orgId) {
+    return NextResponse.json({ status: 'forbidden', message: 'Participant is coded to a different organization' }, { status: 403 });
+  }
+
+  const action = ActivityActions.participantUpdate(
+    params.activityId,
+    participantId,
+    firstname ?? existing?.firstname ?? '',
```
Comment (kohlikohl) [inline src/app/api/partner/v1/activities/[activityId]/participants/route.ts:100]: Should there be validation that `firstname` and `lastname` exists?
These values seem optional in the API and I don't think we want to end up in a situation where a record is created without a name?

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
```diff
@@ -0,0 +1,95 @@
...
+    lastname ?? existing?.lastname ?? '',
+    partner.orgId, // forced to the partner's org — cannot write for other orgs
+    typeof time === 'number' ? time : Date.now(),
+    status,
+    typeof miles === 'number' ? miles : undefined,
+    typeof eta === 'number' ? eta : undefined,
```
Comment (kohlikohl) [inline src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: There should probably be a validation that `eta` is a timestamp in the future.

Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: We are authenticating via our google accounts within our application. And using member DEM numbers as a correlation within the data. Once we get the API hooked up, we'll validate the user and mission IDs work.

For the ETA, yes we have wired up an ETA feature on our application that can also alert our in-town person.

Comment (kohlikohl) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: How would that work if you don't know the participantId yet when a member first signs in?
You wouldn't have a DEM number to compare against in the activity. Or am I missing something?

Also as I think about the API path more, I would advocate that the API is not in a separate partners folder. From an API perspective there should be no difference if the API is used by a partner application or not. Also the activities endpoints already exist and could be extended to check for both auth options as well as taking a scope parameters to check the scope.

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:100]: I had a little more time to look the comments over now that I'm back at my computer. For this one, we assumed all member records are pulled from D4H into the Respond database, so we'd have a record for everyone. I see the concern though. The D4H pull might not run that often, so we could hit situations where there's no record yet. We'll implement a change to require a name when creating a new record.

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: Re: the ETA - Perhaps we can validate that ETA isn't in the past, give or take a few minutes, to allow a small tolerance so that minor clock differences between our app and Respond don't reject a legitimate update. Let me see if we can adjust this in a small change.

Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: Re: the DEM number - Good point on the lack of a DEM connection, especially if the database isn't pulling from D4H often enough. SMR only adds a few new members a few times per year, but if we're drafting a generic API, it should be structured to apply to all cases.

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:57]: My understanding is that `participantId` = `organizationId:memberId`, where the organizationId is a single-number designator for each unit/team and the memberId is pulled from D4H. On the SMR app side, we have some discretion on how we resolve id's. We do have an API into D4H. But have not picked it up at the moment. I was just trying to follow the format that was in Respond in the API and then figure out the downstream connection on our side once we establish the API iteself.

src/app/api/partner/v1/activities/[activityId]/participants/route.ts
Comment (NickConstantine) [reply src/app/api/partner/v1/activities/[activityId]/participants/route.ts:106]: Re: API path - That makes sense to me, and I'm happy to go with your judgment here since you know this codebase way better than I do. I'll be honest, I put it in a separate `partner/` folder mostly because I didn't want to break anything in the existing routes. And it seemed to get us to where we needed to be with access.

The initial thought was to get to an API key that can only access one organization and can't accidentally reach everything a logged-in user can.

Perhaps for now we merged this one with the small fixes first and then if need be expand the API to make it more general at a later point?

PR #169, opened 2026-08-09
fix: sync.ts issue causing dulplicated actions

valonm:
I'm noticing duplicated records when using `addTimeline()` in `DashboardParticipantStatusButton`. It creates 3 entries at a time, all with the same status.

[edit] I'm seeing this issue happen elsewhere, as well. with other commands, e.g. `useTeamCommands();` on `DashboardTeamCard` is causing the same issue when resources are dropped on the team.

```
const participants = useParticipantCommands();
participants.addTimeline(activity.id, participant.id, update);
```

image

After this fix from copilot, it seems to be resolved, but I don't understand the root cause, so I'm not going to merge this yet. When I restart the server (i.e. the state gets cleared and it refreshes from the database) the extra entries disappear -- so it is only persisting once to the database. It has to be something to do with the in memory state on the server; and I think, based on this fix, this means it's related to the websocket.

Based on its root cause analysis I'm wondering if there is more to look at, here. e.g. is this just fixing a symptom.

From Copilot

**Prompt**
it looks like there is only 1 call to the participant command. something else is happening in the state management or perhaps with the websocket?

Updating participant status: {time: 1786254711627, status: 5, organizationId: '1'}
index.ts:14 action cmd/participant/timelineAdd {activityId: 'xxx', participantId: 'xxx', update: {…}}
index.ts:14 action evt/participant/timelineAdded {activityId: 'xxx', participantId: 'xxx', update: {…}}

**Response**
I’m going to trace the command/event flow in the client store and websocket sync path to find where the same participant timeline update is being applied more than once, then I’ll patch the root cause.

I found the event pathway files; now I’m reading the sync implementation and reducers to pinpoint whether events are being applied multiple times or if duplicate server events aren’t being deduped client-side.

The client log shows one command and one event, so the likely issue is reducer behavior for that event; I’m now inspecting the shared participant timelineAdded reducer/mutator to see why one event could append multiple entries.

I found a likely root cause in provider lifecycle: auth/config actions are dispatched during render, which can trigger repeated sync restarts and replayed socket state/events; I’m confirming event shape next, then I’ll patch the provider dispatch pattern to stop reconnect churn.

I’m implementing a sync-layer fix now: restart only when auth actually changes, plus short-window event de-duplication in the websocket handler so repeated server deliveries can’t append timeline entries multiple times.

Nice catch. Your log showing one command plus one participant event strongly points away from the button and toward sync/state flow.

I implemented a different fix in [sync.ts]

Restart socket only when auth actually changes
Before: every AuthActions.set triggered restart.
Now: restart runs only if auth payload meaningfully changed (or logout happened).
De-duplicate repeated websocket events in a short window
Added a 10-second dedupe cache in the sync layer.
Key uses event type + author + timestamp + payload.
If the same stamped event is received again in that window, it is ignored.
Why this helps:

If reconnect churn or websocket replay is causing the same server event to arrive multiple times, the reducer will no longer apply duplicates.
It also reduces unnecessary reconnect cycles that can produce replay-like behavior.

### DIFF
--- a/server/src/stateManager.ts
+++ b/server/src/stateManager.ts
@@ -85,7 +85,6 @@ export class StateManager {
    */
   async handleCommand(command: Command, author: EventAuthor): Promise<void> {
     const events = produceEvents(command);
-    console.log('PROCESSING COMMAND', command.type, events.map((e) => e.type));
     if (!events.length) return;
 
     const stamped: StampedEvent[] = events.map((event) => ({ ...event, meta: { author, timestamp: Date.now() } }));
--- a/server/src/stateManager.ts
+++ b/server/src/stateManager.ts
@@ -85,6 +85,7 @@ export class StateManager {
    */
   async handleCommand(command: Command, author: EventAuthor): Promise<void> {
     const events = produceEvents(command);
+    console.log('PROCESSING COMMAND', command.type, events.map((e) => e.type));
     if (!events.length) return;
 
     const stamped: StampedEvent[] = events.map((event) => ({ ...event, meta: { author, timestamp: Date.now() } }));
--- a/client/src/components/activities/ActivityTile.tsx
+++ b/client/src/components/activities/ActivityTile.tsx
@@ -1,6 +1,7 @@
 import { NearMe } from '@mui/icons-material';
 import { Box, Card, CardActions, Grid, IconButton, Link, Typography } from '@mui/material';
 import { ReactNode } from 'react';
+import { Link as RouterLink } from 'wouter';
 
 import { usePreferences } from '@respond/components/PreferencesProvider';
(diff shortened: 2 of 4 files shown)


server/src/stateManager.ts
```diff
@@ -85,6 +85,7 @@ export class StateManager {
    */
   async handleCommand(command: Command, author: EventAuthor): Promise<void> {
     const events = produceEvents(command);
+    console.log('PROCESSING COMMAND', command.type, events.map((e) => e.type));
```
mcosand: left over debug message

mcosand (approved):
This looks like a much better fix. There's probably work to do on the sync/socket code to more active about preventing duplicate connections, but I don't think it needs to be done now.

